PR #6, opened 2016-02-18
Fixed CSS issues with files. Made website responsive. Removed .idea from gitignore

Raistlfiren:
....


Raistlfiren:
Good call. I have updated my commit to now fix that issue. I also decided to add some news articles and where to add the "Steam" information on the homepage. I had to restructure some of the code for the header.

PR #26, opened 2016-05-18
Miscellaneous Additions

Raistlfiren:
Fixed formatting of navigation.
Updated ENV to development instead of production. 
Updated README to include setting env. 
Added Development dropdown. Fixed responsiveness.
Added font-awesome to project. 
Updated homepage with social media icon.


Raistlfiren:
Fixed conflict by rebasing.

Blodir:
I'm not sure about the development submenu. It makes the navbar very cluttered and doesn't have much content in it. We already have a link to github under contributions (and the footer) and the devblog doesn't need to exist anymore with the new website introducing news/blog categories - so we'll just use that instead. Only thing that we are missing currently is link to the statuspage, I made an issue for it here: #28

Blodir:
Also, the download button in navbar should stay separated from other navigation links to create emphasis

Raistlfiren:
I did notice that the navigation bar is not responsive on tablets. We should get that fixed then. Should we also include Facebook on homepage in footer?

Blodir:
Without the development dropdown the navbar should have enough space on tablets without going to two rows, at least it had when i originally tested it

Blodir:
facebook & other social media would be nice as icons in the footer yes (right now just text)

Blodir:
We could also add them on the "community" section of features if we can make it look nice

Raistlfiren:
Yeah that would be cool. I just added them in the footer under the title "Connect". What other social media outlets do we use than Facebook?

Blodir:
youtube, github

They are already listed under the FAF logo, just change them to icons I guess, don't add another title, it looks much better with a single title on each side of the logo

Issue #40, opened 2016-10-04
Beta Client Download Not Working


jeg623:
When I go to get the Beta version of the client, it gives me a 404.

Raistlfiren:
Thanks @jeg623 for reporting this! :)

PR #41, opened 2016-10-06
Latest client download fix

Raistlfiren:
Fixes issue https://github.com/FAForever/website/issues/40

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,6 @@ public/styles/css/*
 
 #Ignore environment
 .env
+
+#Ignore link.json
+link.json
--- a/crontab
+++ b/crontab
@@ -1,1 +1,2 @@
 0 0 * * * /usr/local/bin/node scripts/extractor.js >> /var/log/cron.log 2>&1
+0 * * * * /usr/local/bin/node scripts/getLatestClientRelease.js >> /var/log/cron.log 2>&1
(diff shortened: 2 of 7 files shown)


Sheeo:
I don't think using an environment variable to store the latest download link is a good idea. This means we need to restart the service every time we change the link?

Sheeo:
As for a better solution--I'm not sure. The database has the latest link within it, but we don't currently have database access from this webservice -- I don't see why not though?

Raistlfiren:
Yeah, I was thinking the same thing @Sheeo. We could just pull it from the database, or I could create a new API endpoint. I think the API endpoint is the best way imho, then we can use it other clients easily.

Maybe I can create that endpoint and another endpoint for creating an account.

Sheeo:
Up to you, I don't mind either way. You should coordinate with @Brutus5000 about account creation.

Unfortunately the PHP-code currently backing the account creation isn't published, you'll have to get in touch with @micheljung to get at it.

Raistlfiren:
@Sheeo I updated this PR to include a static link provided by downlord. This should work the best I hope.

Sheeo:
Static links will cause http cache issues (We've had these in the past, so I'd rather not go back to doing that)

An easy alternative is to use this: https://developer.github.com/v3/repos/releases/#get-the-latest-release

Raistlfiren:
Okay rebased and included a cron job to pull the latest link information from the github api every hour.

PR #48, opened 2016-10-21
Player rating over time Implementation and Leaderboard Updates

Raistlfiren:
- Added rating over time implementation for each user
- Fixed leaderboard ratings to show 1v1 and global
- Added cron job to download all member names every night to populate autocomplete


Raistlfiren:
@Sheeo I think this can be merged as well. Thanks!

Sheeo:
Interesting approach with the cron job :) 👍

PR #106, opened 2017-08-15
Fix api return interpretation

duk3luk3:
api now returns JSON objects where it previously returned a bare 'ok'
string.


Raistlfiren:
Make that one change, and it should be good to go. Thanks sir!

Raistlfiren:
Thank you @duk3luk3!

PR #124, opened 2017-12-23
Update clients comparison

speed2CZ:
Coop is now working in downlord's client.

Since I dont know much about the website stuff, this is just an educated guess to change X to ✓


Brutus5000:
If the criteria are matched by both clients, then we probably need more and/or different criteria. Otherwise the whole comparison is rather useless.

PhilipJFryFAF:
#133

PR #128, opened 2018-01-22
Feat/no cron slim

Sheeo:
Get rid of crontab, use a different image for building vs deploying.


Sheeo:
Was reviewed in private by @duk3luk3.

PR #141, opened 2018-05-18
Use faf-java-api and clean up/reformat code

micheljung:


Brutus5000 (changes requested):
The reset password needs to be considered. We have a semantic change. Against python-api you need to give username and email. In java-api this is merged into one "identifier" (can be either email or username)

https://github.com/FAForever/faf-java-api/blob/develop/src/main/java/com/faforever/api/user/UsersController.java#L91

Description and form fields need to be changed:
https://github.com/FAForever/website/blob/feature/java-api/templates/views/account/resetPassword.pug#L10
API route needs to be changed to camel case and use identifier
https://github.com/FAForever/website/blob/ea6614ce2bdf26ba9ab21c69998cf004ef0ce29b/routes/views/accounts/post/resetPassword.js#L46

Brutus5000 (approved):
From API side everything seems to be migrated now.

Brutus5000:
I have successfully tested registration, username change, email change, password change and steam linking on the test server.

express.js
```diff
@@ -35,17 +35,17 @@ app.use(middleware.clientChecks);
 
 //Set static public directory path
 app.use(express.static('public', {
-    immutable: true,
-    maxAge: 4 * 60 * 60 * 1000 // 4 hours
+	immutable: true,
```
Sheeo, line 38: This whitespace seems to have been exchanged with tabs?

express.js
```diff
@@ -2,16 +2,16 @@
...
+let express = require('express');
 
-var middleware = require('./routes/middleware');
+let middleware = require('./routes/middleware');
 
-var expressValidator = require('express-validator');
-var bodyParser = require('body-parser');
-var passport = require('passport'),
-	OAuthStrategy = require('passport-oauth').OAuth2Strategy;
+let expressValidator = require('express-validator');
```
Sheeo, line 9: Since we're targeting node 9+, it's safe to use `const` -- best do so for `require`.

Sheeo (approved):
Overall looks good, I trust that it works from Brutus' testing.

Moving forward there's definitely more cleanup to do, though. And in general, please do separate whitespace changes and semantic ones into more commits (Even thought github now has an option to ignore it, you'd still like to be able to verify it separately).

PR #147, opened 2018-06-11
Begin work on an age and purchase location survey

aeoncleanse:
This code almost certainly won't work. I've done what I can to start things off and (hopefully) give a fairly clear idea of what I'm going for.

When it comes to storing the data server-side, it is preferable that any data from the survey NOT be linked to the account created. Just have a tally of the options somewhere and increment it. Total anonymity of the data.


Sheeo:
This needs an update of the privacy policy as well.

We can't really get further without means of storing it, though. This either needs to go through the API or another means of talking with our database.

Brutus5000:
Closed due to inactivity

PR #156, opened 2018-07-03
Change text and table header formatting.

germanicianus:
This change makes the client download page non-discriminating.

### DIFF
--- a/public/styles/site/client.sass
+++ b/public/styles/site/client.sass
@@ -25,7 +25,7 @@
 
   col:nth-child(2)
     width: 25%
-    border: 10px solid #59c7fb
+    border: 1px solid #ccc
 
   col:nth-child(3)
     width: 22%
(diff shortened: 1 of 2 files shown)


germanicianus:
The client download page text should be ok, but please double-check the formatting.

1-alex98:
Can you provide an image?

germanicianus:
New text and formatting (updated):
faforever_client_download_page_new

Brutus5000:
The feature comparison table linked is by no means official. This topic is still highly subjective.

germanicianus:
Yes, regarding the linked page it is (the link was already there before), but at least the user isn’t influenced heavily in favour of one client when having his first glance at the page but really needs to have a look at both clients. That we have no „official“ feature list/comparison for the clients is another topic. If desired, I can remove the link to the wiki page, but how should a new user then decide which client to use - by trying them both?

micheljung (approved):

Brutus5000 (approved):

PR #158, opened 2018-07-16
Refactor calendar route

Sheeo:


Brutus5000:
If `return res.render('calendar');` is called in all cases, I'd put it to the end (as it was before).

Sheeo:
That's a hint: It shouldn't, it should be returning a dedicated error view.

As an aside note; early return means you can skip reading the rest of the function to know that the error case is handled at the end of that block, even if it just renders the same template as the other. Subtle repetition (Which can be refactored further), but it saves the reader the effort of having to check the entire else block (and any subsequent code).

Brutus5000:
Obsolete per #238

PR #170, opened 2018-09-15
Registration page: steam linking mandatory warning

Geosearchef:


1-alex98:
Should we not redirect the user to link steam when he registers?

Geosearchef:
This would have to happen on clicking the link in the email, isn't this call just directed to the API? Also forwarding the user to the steam link page won't work as the user will have to login on the website first.

1-alex98:
Hmm yeah, maybe that could be a future thing tho. Because I imagine people will still miss to steam link

micheljung:
I assume most users won't read this but at least we can say "it was there" ;)

PR #174, opened 2018-10-27
Add hint to register retail SC/SC:FA key on Steam for Gold edition.

germanicianus:


Brutus5000:
What's a gold edition? How do I know whether I have one?
Why no just saying: "If you own a retail copy, you can try to activate it on steam. Unfortunately not all retail copies are working on steam" or something like that.

germanicianus:
Register SC **or** SC:FA retail key on Steam -> Get SC Gold Edition (SC **and** SC:FA) on Steam: https://store.steampowered.com/sub/11732/
So maybe my wording is bad. What is your suggestion on wording the above?

germanicianus:
The current wording and formatting/linking is:
>According to [GPG’s statement](https://web.archive.org/web/20140903143132/http://forums.gaspowered.com/viewtopic.php?f=67&t=57459), "ALL existing CD keys for both SC and FA [can be registered] into updated SC:Gold Editions on Steam."

Brutus5000:
Yes, but that is not helping the user the slightest if he still can't link it...

PR #178, opened 2018-11-15
Update FAF Social rules link

EcoNooob:
idk how github works

### DIFF
--- a/templates/views/account/register.pug
+++ b/templates/views/account/register.pug
@@ -12,7 +12,7 @@ block content
           ul.text-left
             li <strong>You may only register one account.</strong> Additional accounts will be banned. If you misbehave using multiple accounts, you may be banned completely.
             li <strong>Your email address is the only way to regain access to your account.</strong> if you lose your password. If you lose both your password and your email, you will lose access to your account.
-            li <strong>Full FAF Social rules:</strong> <a href="https://forums.faforever.com/viewtopic.php?f=2&amp;t=581">FAF Lobby and Forum rules</a>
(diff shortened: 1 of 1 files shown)


Brutus5000 (approved):

templates/views/account/register.pug
```diff
@@ -12,7 +12,7 @@ block content
           ul.text-left
             li <strong>You may only register one account.</strong> Additional accounts will be banned. If you misbehave using multiple accounts, you may be banned completely.
             li <strong>Your email address is the only way to regain access to your account.</strong> if you lose your password. If you lose both your password and your email, you will lose access to your account.
-            li <strong>Full FAF Social rules:</strong> <a href="https://forums.faforever.com/viewtopic.php?f=2&amp;t=581">FAF Lobby and Forum rules</a>
```
Brutus5000, line 15: I don' think the amp; is correct here.

Brutus5000 (changes requested):
Misclicked.

templates/views/account/register.pug
EcoNooob: Yeah I was wondering about that too, but the new link removes that part anyway?

louvenarde: change it, I was about to merge, now i can't

Brutus5000: Ha sorry. I shouldn't check that in a hurry. Yes you removed it. I thought you added it. But I tested it now without the amp; and it works as well.

Brutus5000 (approved):

PR #180, opened 2019-01-01
Removed the legacy client from the website

louvenarde:
Download client is now a direct link to the latest Java client release, the client comparison page is removed.


templates/views/index.pug
```diff
@@ -9,6 +9,10 @@ block content
 				h1 Forged Alliance Forever
 				h3 The Community Driven Lobby for </br> Supreme Commander: Forged Alliance
 			a(class="btn btn-default btn-lg btn-outro btn-danger", href='/client') Download Client
+			p 
+				| Any issue ?
```
germanicianus: Remove blank in front of ‚?‘.

templates/views/index.pug
```diff
@@ -9,6 +9,10 @@ block content
 				h1 Forged Alliance Forever
 				h3 The Community Driven Lobby for </br> Supreme Commander: Forged Alliance
 			a(class="btn btn-default btn-lg btn-outro btn-danger", href='/client') Download Client
+			p 
+				| Any issue ? 
+				a(href=client_download_link, style="color:white;text-decoration:underline;") Try the legacy client
```
germanicianus: Only link ‚legacy client‘, add ‚.‘ at end of sentence.

templates/views/index.pug
louvenarde: sorry, french punctuation :P

templates/views/index.pug
louvenarde: I don't think it's necessary to add a dot, but i'm going to fix the link

germanicianus: Sentences always need punctuation. Don’t you think so?

germanicianus: But don’t include the dot in the link.

louvenarde: Well no not in that case, the sentence is not part of a paragraph, it's floating around, there's no need to end it with a dot if nothing of the same weight is coming after it
Just like there is no dot on the orange "Download client" button

PR #181, opened 2019-02-03
Update linkSteam.pug

Viba:
Website only tells people to have their profile set to public, users also need to have the game details public for steam linking to succeed.
Also added the steam linking guide with pictures as a link to help users.


Rackover:
Tested, works, merging.

PR #186, opened 2019-07-22
Update leagues.json

FtXCommando:
Updating the league to run for August and to create additional divisions to mirror the future division feature.


louvenarde (approved):
Good to go the moment docker builds are fixed

PR #189, opened 2019-08-25
Update dependencies

Brutus5000:


Brutus5000:
Everything seems to work on the test server! Let's bring it to production.

Issue #191, opened 2019-08-27
While registering you get a ✔️ marking while username is taken


jNullj:
I just registered and noticed that the site wouldnt notify me that the username is taken untill i submit my form. Because i started looking for a cool username i had to type the password for each taken username i attempted to register with.
It's even more confusing becouse there is a ✔️ mark that seems to imply the username is avilable.

edit: can be checked with: `https://api.test.faforever.com/data/player?filter[player]=login==%27USERNAME%27`
but for some reason it works for some usernames and not for mine

edit2 : becouse i tried on the testing server, it works perefectly on the main
`https://api.faforever.com/data/player?filter[player]=login==%27USERNAME%27`

i might send a pull request with a fix soon

norraxx:
What is your username in faf?

jNullj:
@norraxx 
you can test with any username, easy to replicate on chrome

im testing a solution but i cant find `account.min.js` anywhere

Brutus5000:
Fixed with #223

PR #192, opened 2019-08-27
quick fix for #191 alreting user of taken username at while creating new account

jNullj:
as a fast fix for Issue #191 i added a simple alert that jumps after the user gets out of focus from the username field, alerting the user if the choosen username is already taken.

Im not fimiliar at all with nodejs web dev, but i hope this fix can help in the meantime and make it easier to create a more elegant solution.

### DIFF
--- a/public/js/app/validation.js
+++ b/public/js/app/validation.js
@@ -1,5 +1,3 @@
-const api_data = "https://api.faforever.com/data/player"
-
 // If username is taken update glyphicon and error message to inform the user
 function renderUserTaken(){
 	// this line shows text in the error area to inform about taken username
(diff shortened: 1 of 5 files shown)


public/js/app/validation.js
```diff
@@ -0,0 +1,4 @@
+$("input[name='username']").change(function(event){$.getJSON("https://api.faforever.com/data/player?filter[player]=login==%27"+ $("input[name='username']").val() +"%27",
+   function(data) {
+     if(!jQuery.isEmptyObject(data.data)){alert("username is already in use");}
+   });})
```
norraxx: Please, format it properly + some comments:
```javascript
$("input[name='username']").change(
	function(event) {
		$.getJSON(
			// https://api.faforever.com/data should be in constants as path
			// getJSON accepts data as second parameter, did you try it?
			"https://api.faforever.com/data/player?filter[player]=login==%27" + $("input[name='username']").val() + "%27",
			function(data) { // i would rename data to result, because it's xhr result object.
				if(!jQuery.isEmptyObject(data.data)) {
					// I think, you can inform user much better, not with alert.
					alert("username is already in use");
				}
			}
		);
	}
);
```

norraxx (changes requested):

public/js/app/validation.js
norraxx: Also on alert you should focus user back to input.

jNullj: updated the file accordingly

1-alex98 (approved):

templates/views/account/register.pug
```diff
@@ -37,4 +37,5 @@ block content
 
 block js
   script(src='/js/account.min.js')
+  script(src='/js/validation.js')
```
janmanfai: Please add the path for the validation.js file to `grunt/concat.js` and remove it from the .pug file. Else it'll throw a 404 error.

jNullj: thank you for pointing out the issue.
recent changes should resolve the issue

public/js/app/validation.js
```diff
@@ -0,0 +1,43 @@
+const api_data = "https://api.faforever.com/data/player"
```
Brutus5000: There is an env variable API_URL pointing to https://api.faforever.com (or the test server, or localhost depending on your environment). This must be used to be environment-agnostic.

public/js/app/validation.js
```diff
@@ -0,0 +1,43 @@
...
+	// this marks the formgroup in red
+	$("input[name='username']").parents("div.form-group").addClass("has-error has-danger");
+}
+
+//clean up username taken message and glyphicon indication for next input attempt
+$("input[name='username']").focus(
+	function(event) {
+		// remove any error relating text, user now tries a new input
+		$("input[name='username']").siblings("div.help-block").eq(1).text("");
+		// dont show glyphicon with ❌ untill checked
```
Brutus5000, line 15: typo: untill

public/js/app/validation.js
```diff
@@ -0,0 +1,43 @@
...
+		$("input[name='username']").siblings("span.glyphicon").removeClass("glyphicon-remove");
+		// remove red markings untill input is checked agian
+		$("input[name='username']").parents("div.form-group").removeClass("has-error has-danger");
+	}
+);
+
+// After user changed the username check if its taken
+$("input[name='username']").change(function(event){
```
Brutus5000: Username must be longer than 3 chars. Also it would be good to have a delay in there (the user has not entered anything for ? seconds). I don't want to have 20 api queries while typing a 20 character long name.

Brutus5000 (changes requested):

public/js/app/validation.js
1-alex98: also this would be a Cross Site Request which is proabably not gonna work

PR #196, opened 2019-09-07
Pre-release ignoring

### LINKED ISSUE #195: Ignore pre-release client releases

1-alex98:
Fixes #195

### DIFF
--- a/express.js
+++ b/express.js
@@ -10,6 +10,7 @@ let expressValidator = require('express-validator');
 let bodyParser = require('body-parser');
 let passport = require('passport');
 let OAuth2Strategy = require('passport-oauth2');
+let schedule = require('node-schedule');
 
 const showdown = require('showdown');
 const fs = require('fs');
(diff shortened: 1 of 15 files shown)


1-alex98:
Currently deployed on test.faforever.com

package.json
```diff
@@ -24,6 +24,7 @@
     "json2csv": "3.7.0",
     "moment": "2.24.0",
     "moment-timezone": "0.5.10",
+    "node-schedule": "^1.3.2",
```
janmanfai: I'd appreciate it if you could provide the corresponding yarn.lock file.

express.js
```diff
@@ -238,23 +239,25 @@ let getLatestClientRelease = require("./scripts/getLatestClientRelease");
 extractor.run();
 getLatestClientRelease.run();
 
-// Run leaderboard extractor every minute
-setTimeout(() => {
+// Run leaderboard extractor every 15 minutes
+schedule.scheduleJob('*/15 * * * *', function myFunc() {
```
janmanfai: An anonymous function is adequate in this case. The same applies to the other scheduleJob.

1-alex98: u mean an arrow function?
I am sorry I got not much of an idea if it comes to javascript?

package.json
1-alex98: hmm, i got massive issue getting this website to run on windows. When I fix some issue the next one pops up...

1-alex98: Maybe I should install me Linux again I remeber it was easy there

express.js
janmanfai: No worries. Just use function() so it looks like this `schedule.scheduleJob('*/15 * * * *', function() {`.

1-alex98: oh ok

1-alex98:
@janteine went for linux now

Brutus5000:
@axel1200 Please resolve the conflict

express.js
```diff
@@ -10,6 +10,7 @@ let expressValidator = require('express-validator');
 let bodyParser = require('body-parser');
 let passport = require('passport');
 let OAuth2Strategy = require('passport-oauth2');
+let schedule = require('node-schedule');
```
louvenarde: Please do not add new dependencies to the website unless you need a new feature. The schedule job change you've added does exactly the same as an interval, except we now have more dependencies.

louvenarde (changes requested):
The update of GetLatestClientRelease.js and middleware.js is good and needed
However I do not see the point of the update of express.js . It is not needed as I see it, and you're adding extra dependencies. Please keep the part of your PR that is about ignoring the pre-release, and drop the part this is about using an extra dependency to do a simple setInterval.

Rackover:
(apart from that the PR is very nice, i do not see much worth nitpicking)

Rackover:
Tidied it up to keep only the release ignoring, merging in

louvenarde (approved):

PR #197, opened 2019-09-19
Live player count 

louvenarde:
Displays live player count and game count on homepage, taken from python server API

### DIFF
--- a/public/js/app/counter.js
+++ b/public/js/app/counter.js
@@ -1,0 +1,26 @@
+const delay = 3000;
+
+$("#game_counter").parent().hide().fadeIn();
+            
+setInterval(function(){
+    $.get('lobby_api', { resource: "players"}, function (body) {
+        let entries = body;
+        if ($("#player_counter").text() == entries.length.toString()) return;
+        $("#player_counter").fadeOut('fast', function() {
+            $(this).text(entries.length).fadeIn('fast');
+        })
+    });
+}, delay);
+
+setTimeout(() => {
+    setInterval(function(){
+        console.log($("#game_counter").parent());
+        $.get('lobby_api', { resource: "games"}, function (body) {
+            let entries = body;
+            if ($("#game_counter").text() == entries.length.toString()) return;
+            $("#game_counter").fadeOut('fast', function() {
+                $(this).text(entries.length).fadeIn('fast');
+            })
+        });
+    }, delay);  
+}, delay/2);
--- a/public/js/app/counter.js
+++ b/public/js/app/counter.js
@@ -1,8 +1,24 @@
 const delay = 1000;
-
-$("#game_counter").parent().hide().fadeIn();
             
-setInterval(function(){
+setInterval(updatePlayerCounter, delay);
+
+setTimeout(function(){
+    setInterval(updateGameCounter, delay);  
+}, delay/2);
+
+function updateGameCounter(){
+    $.get('lobby_api', { resource: "games"}, function (body) {
+        let games = body;
+        
+        if ($("#game_counter").text() == games.length.toString()) return;
+        
+        $("#game_counter").fadeOut('fast', function() {
+            $(this).text(games.length).fadeIn('fast');
+        })
+    });
+}
+
+function updatePlayerCounter(){
     $.get('lobby_api', { resource: "players"}, function (body) {
         let players = body;
         if ($("#player_counter").text() == players.length.toString()) return;
@@ -10,18 +26,17 @@ setInterval(function(){
             $(this).text(players.length).fadeIn('fast');
         })
     });
-}, delay);
+}
 
-setTimeout(function(){
-    setInterval(function(){
-        $.get('lobby_api', { resource: "games"}, function (body) {
-            let games = body;
-            
-            if ($("#game_counter").text() == games.length.toString()) return;
-            
-            $("#game_counter").fadeOut('fast', function() {
-                $(this).text(games.length).fadeIn('fast');
-            })
-        });
-    }, delay);  
-}, delay/2);
+updateGameCounter();
+updatePlayerCounter();
+
+const waitBeforeShowing = setInterval(
+    function(){
+        if ($("#player_counter").text() != '0' && $("#game_counter").text() != '0'){
+            $(".counters").css("opacity", 1);
+            clearInterval(waitBeforeShowing);
+        }
+    },
+    100
+);
--- a/public/js/app/counter.js
+++ b/public/js/app/counter.js
@@ -1,25 +1,26 @@
-const delay = 3000;
+const delay = 1000;
 
 $("#game_counter").parent().hide().fadeIn();
             
 setInterval(function(){
     $.get('lobby_api', { resource: "players"}, function (body) {
-        let entries = body;
-        if ($("#player_counter").text() == entries.length.toString()) return;
+        let players = body;
+        if ($("#player_counter").text() == players.length.toString()) return;
         $("#player_counter").fadeOut('fast', function() {
-            $(this).text(entries.length).fadeIn('fast');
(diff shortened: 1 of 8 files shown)


routes/views/index.js
```diff
@@ -6,7 +6,7 @@ exports = module.exports = function(req, res) {
   // locals.section is used to set the currently selected
   // item in the header navigation.
   locals.section = 'home';
-
+
```
norraxx, line 9: :monkey:

public/js/app/counter.js
```diff
@@ -0,0 +1,26 @@
+const delay = 3000;
+
+$("#game_counter").parent().hide().fadeIn();
+            
+setInterval(function(){
+    $.get('lobby_api', { resource: "players"}, function (body) {
+        let entries = body;
+        if ($("#player_counter").text() == entries.length.toString()) return;
```
norraxx: Please, write `{}` for simple `return`. Now how it is is very ugly to read.

public/js/app/counter.js
```diff
@@ -0,0 +1,26 @@
...
+        $("#player_counter").fadeOut('fast', function() {
+            $(this).text(entries.length).fadeIn('fast');
+        })
+    });
+}, delay);
+
+setTimeout(() => {
+    setInterval(function(){
+        console.log($("#game_counter").parent());
```
norraxx: Don't leave console.log

public/js/app/counter.js
```diff
@@ -0,0 +1,26 @@
+const delay = 3000;
+
+$("#game_counter").parent().hide().fadeIn();
+            
+setInterval(function(){
+    $.get('lobby_api', { resource: "players"}, function (body) {
```
norraxx, line 22:
```
setInterval(function() {
    $.get('lobby_api', {resource: "players"}, function (body) {
```

Well, you're mixing `() => {}` and `function () {}`. Choose one of this.

public/js/app/counter.js
```diff
@@ -0,0 +1,26 @@
...
+        })
+    });
+}, delay);
+
+setTimeout(() => {
+    setInterval(function(){
+        console.log($("#game_counter").parent());
+        $.get('lobby_api', { resource: "games"}, function (body) {
+            let entries = body;
+            if ($("#game_counter").text() == entries.length.toString()) return;
```
norraxx: -//-

public/js/app/counter.js
```diff
@@ -0,0 +1,26 @@
...
+        $("#player_counter").fadeOut('fast', function() {
+            $(this).text(entries.length).fadeIn('fast');
+        })
+    });
+}, delay);
+
+setTimeout(() => {
+    setInterval(function(){
+        console.log($("#game_counter").parent());
+        $.get('lobby_api', { resource: "games"}, function (body) {
```
norraxx: same, as above, format it well. ;-) + ` setInterval(function(){` too

public/js/app/counter.js
```diff
@@ -0,0 +1,26 @@
...
+            let entries = body;
+            if ($("#game_counter").text() == entries.length.toString()) return;
+            $("#game_counter").fadeOut('fast', function() {
+                $(this).text(entries.length).fadeIn('fast');
+            })
+        });
+    }, delay);  
+}, delay/2);
```
norraxx: \n

public/js/app/counter.js
```diff
@@ -0,0 +1,26 @@
+const delay = 3000;
```
norraxx: But in fact, I **really** don't like to ask database twice every 3 seconds. Make it once per 30 seconds.
Don't forget, that database is our big bottleneck, we prefer to not to make any unnecessary load on it.

public/styles/site/index.sass
```diff
@@ -109,3 +109,6 @@
 
 .btn-outro
   margin: 30px
+
+.counter
```
norraxx: Do you need that class defined in `sass`?

routes/lobby_api.js
```diff
@@ -0,0 +1,6 @@
+exports = module.exports = function(req, res) {
+    const request = require('request');
+    request(process.env.LOBBY_API_URL + '/'+req.query.resource, function (error, response, body) {
```
norraxx: `request(process.env.LOBBY_API_URL + '/' + req.query.resource, function (error, response, body) {`

public/js/app/counter.js
```diff
@@ -0,0 +1,26 @@
...
+
+$("#game_counter").parent().hide().fadeIn();
+            
+setInterval(function(){
+    $.get('lobby_api', { resource: "players"}, function (body) {
+        let entries = body;
```
norraxx: body could be named as `entries` and this line can be deleted.
I think this variable should be named `players`

public/js/app/counter.js
```diff
@@ -0,0 +1,26 @@
...
+            $(this).text(entries.length).fadeIn('fast');
+        })
+    });
+}, delay);
+
+setTimeout(() => {
+    setInterval(function(){
+        console.log($("#game_counter").parent());
+        $.get('lobby_api', { resource: "games"}, function (body) {
+            let entries = body;
```
norraxx: `body` can be named as entries and this line can go.
I think, this variable should be named as `games`.

routes/lobby_api.js
```diff
@@ -0,0 +1,6 @@
...
+        res.send(JSON.parse(body));
+    });
+};
```
norraxx, line 6: Do I understand right, that we're asking our website to ask the api, not the api directly?
Is there any reason to create such unnecessary load?

norraxx (changes requested):

norraxx (changes requested):
Well, i think, i think... you can create cache system, ask player count on server side, store result to cache, and read it from cache to outside world.
1. less queries to db
2. you'll have data already from cache and don't have to show 0 after opening the page.

routes/lobby_api.js
louvenarde: CORS Policy

public/js/app/counter.js
Rackover: > Well, i think, i think... you can create cache system, ask player count on server side, store result to cache, and read it from cache to outside world.
> 
> 1. less queries to db
> 2. you'll have data already from cache and don't have to show 0 after opening the page.

We don't want caching for this, because we want the player count to be LIVE on the webpage!
Unless we fake it by adding random deviations to a cached number, but this will be easy to tell apart...

Also, no database is involved in that query. It's using the python server API.

Rackover:
> Well, i think, i think... you can create cache system, ask player count on server side, store result to cache, and read it from cache to outside world.
> 
> 1. less queries to db
> 2. you'll have data already from cache and don't have to show 0 after opening the page.

We don't want caching for this, because we want the player count to be LIVE on the webpage!
Unless we fake it by adding random deviations to a cached number, but this will be easy to tell apart...

Also, no database is involved in that query. It's using the python server API.

Rackover:
The "show 0" thing I'm going to remove, as I'm said on zulip i'm not quite done yet. Thats why there is a unused class in sass. For now.

routes/views/index.js
louvenarde: ???

routes/lobby_api.js
louvenarde: ???

Brutus5000: I guess this is about the inconsistent use of whitespace inside the method

public/js/app/counter.js
louvenarde: (as discussed in zulip this is N/A )

routes/lobby_api.js
louvenarde: Thanks!
Fixed

PR #198, opened 2019-09-19
Awesomium now has a limit of 5 instead of 15, making it slightly fast…

louvenarde:
…er. Also allows to search for 1 characters nicknames.


public/js/app/leaderboards.js
```diff
@@ -113,7 +113,9 @@ var searchbar = document.getElementById("searchbar");
 
 // Show label but insert value into the input:
 new Awesomplete(searchbar, {
-  list: JSON.parse(members)
+    list: JSON.parse(members),
+	minChars: 0,
+	maxItems: 5
```
norraxx: formatting :-D

norraxx (changes requested):

PR #199, opened 2019-09-19
Report form improvements

louvenarde:
Report form is now much easier to use, hints you the name of people, allows to easily add new people to your report, etc


scripts/getRecentUsers.js
```diff
@@ -0,0 +1,46 @@
...
+                    if (error) {
+                        console.log(error);
+                    } else {
+                        console.log(moment().format("DD-MM-YYYY - HH:mm:ss") + ' - Recent players file created successfully');
+                    }
+                });
+
+            }
+        }
+    );
```
norraxx, line 49: extract response callback

scripts/getRecentUsers.js
```diff
@@ -0,0 +1,46 @@
...
+        ;
+     
+	request(
+        requestUrl,        
+        function (error, response, body) {
+            if (! error) {
+                let entries = JSON.parse(body);
+                let finalArray = [];
+
+                if (entries.included == undefined) return;
```
norraxx: add brackets, please

scripts/getRecentUsers.js
```diff
@@ -0,0 +1,46 @@
...
+            '&fields[player]=login'+
+            '&sort=-endTime'
+        ;
+     
+	request(
+        requestUrl,        
+        function (error, response, body) {
+            if (! error) {
+                let entries = JSON.parse(body);
+                let finalArray = [];
```
norraxx: i don't like naming `entries` and `finalArray`, i wish it would be named better.

scripts/getRecentUsers.js
```diff
@@ -0,0 +1,46 @@
...
+
+                if (entries.included == undefined) return;
+
+                for(let i = 0; i < entries.included.length; i++) {
+                    let entry = entries.included[i];
+                    if (entry.type != "player") continue;
```
norraxx, line 34: my eyes, please no

express.js
```diff
@@ -247,6 +249,15 @@ setTimeout(() => {
...
 }, 60 * 1000);
 
+// Run recent players detection every 15 minutes
+setTimeout(() => {
+	try {
+        getRecentUsers.run();
+	} catch (e) {
+		console.error("Error while updating recent user list!", e);
+	}
+}, 15 * 60 * 1000);
```
norraxx: formatting

norraxx: move 15 minutes to constant

scripts/getRecentUsers.js
```diff
@@ -0,0 +1,46 @@
...
+                        console.log(error);
+                    } else {
+                        console.log(moment().format("DD-MM-YYYY - HH:mm:ss") + ' - Recent players file created successfully');
+                    }
+                });
+
+            }
+        }
+    );
+};
```
norraxx: \n

scripts/getRecentUsers.js
```diff
@@ -0,0 +1,46 @@
...
+                if (entries.included == undefined) return;
+
+                for(let i = 0; i < entries.included.length; i++) {
+                    let entry = entries.included[i];
+                    if (entry.type != "player") continue;
+
+                    finalArray.push(entry.attribute.login);
+                }
+
+                fs.writeFile("members/recent.json", JSON.stringify(finalArray), function(error) {
```
norraxx: Do you always have right to write on filesystem? Is website running as root or directory owner?

norraxx (changes requested):
I think, you've created file storage fabrique instead of server side caching.
Maybe here is the example how to create the simpliest cache https://www.compose.com/articles/api-caching-with-redis-and-nodejs/

scripts/getRecentUsers.js
louvenarde: Yes it is, that's how caching is done for the leagues and leaderboards, so I'm merely copying what they do so we don't have 3 different caching systems for 3 systems

scripts/getRecentUsers.js
louvenarde: I do not understand

express.js
```diff
@@ -245,7 +247,16 @@ setTimeout(() => {
 	} catch (e) {
 		console.error("Error while updating leaderboards!", e);
 	}
-}, 60 * 1000);
+},  parseInt(process.env.LEADERBOARDS_UPDATE_INTERVAL) * 1000);
+
+// Run recent players detection every 15 minutes
+setTimeout(() => {
```
Brutus5000: Please check #196 as axel1200 fixes the broken intervalls.

routes/views/accounts/get/report.js
```diff
@@ -93,7 +101,15 @@ exports = module.exports = function(req, res) {
                     'statusStyle':statusStyle
                 });
             }
-            res.render('account/report', {flash: flash});
+            
+            fs.readFile('members/recent.json', 'utf8', function (err, data) {
+                try{
+                    locals.reportable_members = JSON.parse(data);
+                }
+                catch{}
```
Brutus5000: Never empty catch. At least log something.

routes/views/accounts/post/report.js
```diff
@@ -22,7 +22,11 @@ exports = module.exports = async function (req, res) {
   let overallRes = res;
 
   // validate the input
-  req.checkBody('offenders', 'Please indicate the player or players you\'re reporting').notEmpty();
+  let i = 0;
+  while(req.body["offender_"+i]){
+    req.checkBody("offender_"+i, 'Please indicate the player or players you\'re reporting').notEmpty();
```
Brutus5000: Why not use quotes consistently, then you don't have to escape the apostrophe?

routes/views/accounts/post/report.js
```diff
@@ -22,7 +22,11 @@ exports = module.exports = async function (req, res) {
   let overallRes = res;
 
   // validate the input
-  req.checkBody('offenders', 'Please indicate the player or players you\'re reporting').notEmpty();
+  let i = 0;
+  while(req.body["offender_"+i]){
```
Brutus5000: I'm no JS expert, but wouldn't it be less ugly to do something like req.body.filter(startswith("offender_").foreach() ? i-index and while-loop looks so 1980 😃

routes/views/accounts/post/report.js
```diff
@@ -48,7 +52,14 @@ exports = module.exports = async function (req, res) {
 
     const isGameReport = req.body.game_id != null;
 
-    const offenders = req.body.offenders.split(' ');
+    let offenders = [];
+    let j = 0;
+    while(req.body["offender_"+j]){
```
Brutus5000: same as above

scripts/getRecentUsers.js
```diff
@@ -0,0 +1,46 @@
...
+	request(
+        requestUrl,        
+        function (error, response, body) {
+            if (! error) {
+                let apiRecentPlayers = JSON.parse(body);
+                let recentPlayers = [];
+
+                if (apiRecentPlayers.included == undefined) return;
+
+                for(let i = 0; i < apiRecentPlayers.included.length; i++) {
```
Brutus5000: apiRecentPlayer.foreach() ?

Brutus5000 (changes requested):

express.js
louvenarde: He is adding a new useless dependency to do something that JS already does, and he doesnt move the intervals to constant, so I do not see the point.

routes/views/accounts/post/report.js
louvenarde: I don't think I can get the i with a foreach. I've fiddled around a bit and For does what i want now :)

scripts/getRecentUsers.js
louvenarde: `for` is used everywhere so I'm going to keep it. However I can certainly improve it yes

express.js
Brutus5000: The point is the intervals do not work in the current website and nobody so far could tell my why.

louvenarde: Then this is a problem to look into. `setInterval` is, according to w3c docs, not supposed to be broken. Adding a new dependency ontop of it without understanding the underlying problem is not the best of ideas.

louvenarde: I'll look into it once i'm done with this PR and the other

PR #209, opened 2019-11-09
Added button to create account using an account creation token

louvenarde:


Rackover:
Untested as of now

Rackover:
Works on test

Brutus5000 (approved):
Tested it on dev successfully end-2-end

PR #217, opened 2020-01-16
Take care when api is down, site will survive

norraxx:
Well, we will stop when lobby will restart.


Brutus5000 (approved):

Brutus5000:
Should also check the status code etc, but is ok for a fast fix.

PR #218, opened 2020-01-21
Fix favorite faction display

Syldarion:
These changes include a fix to the favorite faction display and an update to the formatting of the changed file. I'll remove the latter if it isn't wanted, it was just to make working on it easier.

The favorite faction fix is explained in the second commit message.

### DIFF
--- a/scripts/updateLeagues.js
+++ b/scripts/updateLeagues.js
@@ -59,12 +59,31 @@ module.exports.run = async function run(leagueData) {
                     }
                 }
 
+                if (leagueData.games == undefined) leagueData.games = [];
+                if (leagueData.playerData == undefined) leagueData.playerData = [];
+
                 // Parsing player stats
                 for (let k in entries.included) {
                     const record = entries.included[k];
                     switch (record.type) {
                         case "gamePlayerStats":
                             const pid = record.relationships.player.data.id;
+
+                            // Let's check if we don't have that player already
+                            let already = false;
+                            for (category in leagueData.playerData) {
+                                if (!already) {
+                                    for (playerIndex in leagueData.playerData[category].data) {
+                                        const registeredPlayer = leagueData.playerData[category].data[playerIndex];
+                                        if (registeredPlayer.id == pid) {
+                                            players[pid] = registeredPlayer;
+                                            already = true;
+                                            break;
+                                        }
+                                    }
+                                }
+                            }
+
                             if (record.attributes.score > 0) {
                                 games[record.relationships.game.data.id].winner = pid;
                             }
@@ -92,9 +111,6 @@ module.exports.run = async function run(leagueData) {
                             break;
                     }
                 }
-
-                if (leagueData.games == undefined) leagueData.games = [];
-                if (leagueData.playerData == undefined) leagueData.playerData = [];
 
                 let categories = {};
 
--- a/scripts/updateLeagues.js
+++ b/scripts/updateLeagues.js
@@ -53,7 +53,7 @@ module.exports.run = async function run(leagueData) {
                     const record = entries.data[k];
                     games[record.id] = {
                         'winner': null,
-                        'participants': [],
+                        'participants': {},
                         'startTimestamp': moment(record.attributes.startTime).unix(),
                         'endTimestamp': moment(record.attributes.endTime).unix()
                     }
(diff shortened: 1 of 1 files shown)


Brutus5000 (changes requested):
Generally I don't have a problem with your code, but given the fact that we recently spent multiple hours debugging performance issues of this application I cannot accept 3 nested for loops inside a while loop.

We need to find better data structures for this.

Syldarion:
For sure, I'll take another stab at this then.

norraxx:
@Syldarion Maybe, we could talk about this code at discord. You would explain what you want to achieve, and I'd say what I don't like on this commit, what do you think?

Syldarion:
@norraxx Basically, just trying to have favorite faction percentage based off all games rather than those found first. I can see how just re-using a nested for loop is undesirable, seeing as it'll be going through the same data and processing it the same way twice.

norraxx:
@Syldarion Imagine how much load you have to do, to just compute that stuff. As I said, let's talk about that problem on discord https://discord.gg/VYX25M.
Write me there, i'm `dragonite` user there, maybe we can figure how to make it better. As @Brutus5000 said your solution will make very big load on servers, web page will start minutes, and maybe you won't have enough memory, and what you want is compute 4 numbers. :-)

scripts/updateLeagues.js
```diff
@@ -5,244 +5,260 @@ let fs = require('fs');
...
+                        'winner': null,
+                        'participants': [],
+                        'startTimestamp': moment(record.attributes.startTime).unix(),
+                        'endTimestamp': moment(record.attributes.endTime).unix()
                     }
                 }
-                
+
+                if (leagueData.games == undefined) leagueData.games = [];
+                if (leagueData.playerData == undefined) leagueData.playerData = [];
```
norraxx, line 63: I don't like conditions like that, use brackets, it's ok to have 3 lines instead of 1.

Syldarion:
@norraxx I spent some time trying to come up with a better solution, but only managed to find either longer runtime or wrong data.
I'd ask that you try the change yourself. Locally, running with this change is within a couple seconds of the runtime of what is live, but comes with accurate data.
Or I could close this and open an issue.

Syldarion:
As mentioned in the commit message of what I just added, an obvious solution had been staring me in the face all along. The for loop I added previously is gone, and performance seems in line with live.

norraxx:
I would to meet on discord.

scripts/updateLeagues.js
```diff
@@ -5,244 +5,251 @@ let fs = require('fs');
...
+                let categories = {};
+
+                for (let k in games) {
+                    if (leagueData.games.indexOf(k) > -1) {
+                        continue;   // Game has already been calculated, skipping it
+                    }
+                    const game = games[k];
+                    const timeElapsed = game.endTimestamp - game.startTimestamp;
+
+                    for (l in game.participants) {
```
norraxx, line 108: use let

scripts/updateLeagues.js
```diff
@@ -5,244 +5,251 @@ let fs = require('fs');
...
+                            'rating': playerRecord.rating,
+                            'points': 0,
+                            'factions': playerRecord.factions,
+                            'wld': { "w": 0, "l": 0, "d": 0 },
+                            'secondsPlayed': timeElapsed
+                        };
+
+                        // Let's check if we don't have that player already
+                        let already = false;
+                        for (category in leagueData.playerData) {
```
norraxx, line 123: -//-

scripts/updateLeagues.js
```diff
@@ -5,244 +5,251 @@ let fs = require('fs');
...
-                            if (!already){
-                                const catName = getRangeName(leaguesConfig.ratingRange, player.rating);
-                                let found = false;
-                                for (categoryIndex in leagueData.playerData){
-                                    if (leagueData.playerData[categoryIndex].categoryName == catName){
-                                        leagueData.playerData[categoryIndex].data.push(player);
-                                        found = true;
+                }
+
+                let categories = {};
```
norraxx, line 99: you don't need it

norraxx (commented):
Cosmetic changes.

Brutus5000 (approved):

PR #224, opened 2020-02-25
Update leagues.json

FtXCommando:


1-alex98:
That's though man  this means be also need to put that in producton by 01/03/2020

1-alex98 (approved):

1-alex98:
@Brutus5000 u need to merge this because u need to also deploy it before 01/03 otherwise we need to change date

Brutus5000 (approved):

PR #225, opened 2020-02-25
Add information about limitation of user name change

### LINKED ISSUE #188: Add information about limitation of user name change
We need to tell people how often they can change their username and for how long the old name will be reserved.

1-alex98:
Fixes #188

### DIFF
--- a/templates/views/account/changeUsername.pug
+++ b/templates/views/account/changeUsername.pug
@@ -8,7 +8,7 @@ block content
             .col-md-12
                 h1.account-title Change Username
                 h4.account-subtitle.text-center Change your username by using the form below. Once it is changed, you can log in to your account with the new username.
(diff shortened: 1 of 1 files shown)


templates/views/account/changeUsername.pug
```diff
@@ -8,6 +8,7 @@ block content
             .col-md-12
                 h1.account-title Change Username
                 h4.account-subtitle.text-center Change your username by using the form below. Once it is changed, you can log in to your account with the new username.
+                h4.account-subtitle.text-center You can only change the name every 30 days and your old username will be reserved for 6 months. In case you choose an offending username you might get banned for 30 days.
```
Brutus5000: We don't have exact rules about banning periods. Just write you might get banned

Brutus5000 (changes requested):

Brutus5000 (approved):

PR #246, opened 2020-06-28
Update leagues.json

FtXCommando:


1-alex98 (approved):

1-alex98:
We really need to build something so this is does not require a new website release

Brutus5000:
We really need to get rid of this, as it makes the api query like 100k players and rating with SINGLE queries.

1-alex98:
oh yeah that as well

1-alex98:
Rather 10k players and their games right

Brutus5000:
I just saw an endless log stream on test (hibernate sql log is active there)

PR #247, opened 2020-07-03
Newshub open links in new tab

Askaholic:
I feel like this would be the expected behavior. Would also be nice if any links within the paragraph also got the `target="_blank"` attribute, but I have no idea how to do that with pug.


1-alex98 (approved):

Brutus5000 (approved):
Tested on test and works

PR #259, opened 2021-01-18
Add Twitch link to Top Navigation Bar

MrRowey:
Added Link to the FAFLive Twitch Account along side the Discord and Youtube Link


templates/layouts/default.pug
```diff
@@ -45,6 +45,8 @@ html
                 i.icon.fab.fa-discord
               li: a.icon(href="https://www.youtube.com/c/ForgedAllianceForever")
                 i.fab.fa-youtube
+              li: a.icon(href="https://www.twitch.tv/faflive")
```
Brutus5000: faflive is offline 99% of the time. Wouldn't it be better to link to the FA category of twitch?

PR #270, opened 2021-04-22
Update lobby_api.js

Kazbek:
Trying to cache full response of lobby games api.


1-alex98 (approved):
Correct indenting would be good

Kazbek:
Ping me if something needs to be changed/fixed.

PR #280, opened 2021-05-21
Livestream Tab with Embeded Iframe for FAFLive

MrRowey:
The Iframe Loads and it pointing to the correct location. the only thing I cannot test locally is if the website is allowing it to display the stream due to Twitch Requires SSL webpage to embed which my Local version of the Site Dose not have. If the Follwoing changes dont let it Though thier are addition line i will need to add to the Express.js & Package.json


MrRowey:
Do not Merge this Currenly :)

PR #283, opened 2021-07-06
Remove username check while public api is down

Sheikah45:
Not sure if there is some other public api dependent function behind this


1-alex98 (approved):

1-alex98:
Just wanted to do the same

PR #296, opened 2021-09-16
Removed top 5 players displaying in site information

MrRowey:
Removed the lines that displayed the top 5 players names with their true skill number


MrRowey:
- Merged in with 2v2 Leaderboard support

PR #302, opened 2021-10-25
Create Headroom only when header exists

### LINKED ISSUE #284: Uncaught TypeError when loading website without header
setting User Agent to remove header and footer causes `Uncaught TypeError: Cannot read property 'classList' of null`.
probably because of this:
https://github.com/FAForever/website/blob/b6c833156108fd40aff8b34ff2d547d53d95773c/public/js/app/navigation.js

`document.querySelector('.navigation-wrapper')` results in `null` and then calling `headroom.init()` gives this error.

Gatsik:
Resolves #284


Brutus5000 (approved):
LGTM

PR #310, opened 2021-12-18
Add Dart SASS

MrRowey:
As Discussed in zulip Due to the docker build using the latest LTS version of the node the Node-SASS module that is depreciated does not play well with it so will need to remove node-sass and replace with Dart-SASS


MrRowey:
oh, fun with setting dart as the new sass interpreter the file that used semi-colons at the end of each sass line is no longer happy. will need to remove the semi colons and see if it happy with that change afterwards CSS/sass doesn't use them so not sure why their their

PR #311, opened 2021-12-19
Website Updates

MrRowey:
This is the first set of improvements to reduce the load times for the website.

- Replaced old images with more up-to-date ones
- Changed Majority of Images from PNG to JPG while still retaining the Quality
- Removed Globe on the page due to added unnecessary time on first load time with it spamming update requests


templates/views/index.pug
```diff
@@ -95,5 +88,3 @@ block content
 block js
   script const refreshCountersSeconds = !{refreshCountersSeconds};
   script(src="/js/app/counter.js")
-  script(src="/js/world-map.min.js")
-  script(src="/js/world-map-dependencies.min.js")
```
Brutus5000, line 99: As far as I know there is a whole lot of CSS that needs to be deleted in addition to the javascript.

MrRowey:
Closing in favour for #313 for fixing something I broke but could not see what in this PR

PR #365, opened 2022-04-01
Fix player count

louvenarde:


MrRowey (approved):

MrRowey:
image

Rackover:
I thought this had been tested before merging. Why was this merged, and closed, if it doesn't work ? :( 
It indeed references the flash code that I did not include here because I wasn't sure if it was needed, the bug is on me. Since this PR has been merged, I will create another one. Please test the PR **before** merging next time 😅 testing after merging instead of before creates more work for everyone

Issue #326, opened 2022-01-16
Active Players Stats on Landing page


MrRowey:
The stats showing the active games and and how many people are online has disappeared in changes that I have made that I did not notice. need to revisit all the changes to see where I remove this

Rackover:
![image](https://user-images.githubusercontent.com/33836535/161329160-a0bf6a79-90ce-4489-b170-eb175c079687.png)

![image](https://user-images.githubusercontent.com/33836535/161329763-355023f3-1d30-4675-8e48-8b30e6819759.png)

![image](https://user-images.githubusercontent.com/33836535/161329779-78d51157-7238-4cae-bfaa-04bdfb40b64c.png)

Interpolation fails because this critical setting is missing: 
https://github.com/FAForever/website/blame/b7a85dca5dd8ca31bbc643c9c44ab8359a0279e3/routes/views/index.js#L33

Has been removed by this commit:
https://github.com/FAForever/website/commit/b58580008b3dc19f60c0b364e127049b09a6badb

Along with a few other things that I think are essential (the flash errors you removed are used to display login error & clan app errors if I remember correctly)

You can fix this by simply adding this back
```
    // Render the view
    res.render('index', {flash: flash, refreshCountersSeconds: parseInt(process.env.PLAYER_COUNT_UPDATE_INTERVAL)});
```

but unless the flash errors have become unused throughout the whole website (something to check for thoroughly) then yo should probably also add them back because they will cause other issues elsewhere

```
    let flash = {};

    if (req.query.flash) {
        let buff = Buffer.from(req.query.flash, 'base64');  
        let text = buff.toString('ascii');

        try {
            flash = JSON.parse(text);
        } catch(e) {
          console.error("Parsing error while trying to decode a flash error: " + text);
          console.error(e);
          flash = [{msg: "Unknown error"}];
        }
    }
```

from where it was taken (index.js in website\routes\views)

Look at git blame on index.js to check the previous file structure & where to put the things

and you should be good for a test & deploy 🙂

JaviTrek:
Thanks for the response @Rackover , I've been toying with the new frontend for the website and was wondering what were the flash messages for lol. Thanks alot for the detailled answer!

MrRowey:
[...]

PR #367, opened 2022-04-03
Fix player count & restore flash system on homepage

louvenarde:
**make sure to test this locally before merging**, and **never merge a PR that has not been reviewed** ! 🙏 

There are people among faf who can read javascript, hand it over to them and **wait for the code to be reviewed** before merging anything!

(this is true of all code, not just this PR ♥ )


MrRowey (approved):
Checked with Deployment with test sever and is not throwing the error that was previously being shown here #326

PR #382, opened 2022-04-11
1V1

JaviTrek:
I made a mess with my branch management and now have this abomination, would try to fix this but I'm worried I'll end up messing up my new code so I pray this isn't too hard or idk lets do voice and see if you can pull your magic @MrRowey . Also github is god awful with the whole process needed to just update stuff.


MrRowey (commented):
members should have a gitignore and have no file in it as thet are generate on the webtie startup

MrRowey:
image
those 2 files need to be deleted / add a git ignore to the folder so they dont get merge in

MrRowey:
will merge with branch one they have been delt with :)

PR #386, opened 2022-05-29
Web Site Minor Clean Up

MrRowey:
-  Removed Tournaments.pug Page and all corresponding pages due to its main use which displayed the FAF Challonge did not work and the Challonge API was causing long load times.

- Update Competitive tab to link to the Global ranks & added in 4v4 Full share

- Minor Page Clean ups

- Updated Main page and added GOG buy Link

- Updated Account Register due to still having Steam link mandatory.


routes/views/4v4.js
```diff
@@ -0,0 +1,23 @@
...
+	// locals.section is used to set the currently selected
+	// item in the header navigation.
+	locals.section = 'competitive';
+	locals.cSection = '4v4';
+	locals.ratingTypeTitle = '4v4';
+	locals.ratingType = 'tmm_4v4_full_share';
+	locals.apiURL = process.env.API_URL;
+
+    fs.readFile('members/4v4.json', 'utf8', function (err, data) {
+        if(err) return;
```
louvenarde, line 15: Log error when it happens, do not fail silently 🙂

routes/views/4v4.js
```diff
@@ -0,0 +1,23 @@
...
+        locals.members = JSON.parse(data);
+        locals.lastPage = Math.ceil(locals.members.length / 100);
```
louvenarde, line 17: `local.members` not checked, what if JSON.parse fails?

louvenarde, line 17: 100 needs to be a variable / constant defined somewhere, no magic/absolutes (it's the page size)

scripts/extractor.js
```diff
@@ -91,9 +92,25 @@ module.exports.run = function run() {
...
+    request(process.env.API_URL + "/data/leaderboardRating?include=player&sort=-rating&filter=leaderboard.id==4;updateTime=ge=" +
+      pastMonth.format("YYYY-MM-DDTHH:mm:ss") + "Z", function (error, response, body) {
+      if (error || response.statusCode > 210) {
+        console.log(moment().format("DD-MM-YYYY - HH:mm:ss") + ' - There was an issue while fetching leaderboards 4v4:');
+        console.error(error);
+        if (response) console.trace(response.statusCode);
+        return;
+      }
+
+      const ratings = models.sync(JSON.parse(body));
```
louvenarde, line 107: Check JSON.parse or wrap it in a try catch

scripts/extractor.js
```diff
@@ -91,9 +92,25 @@ module.exports.run = function run() {
 
       saveLeaderboardRatingsToFile(ratings, "members/2v2.json");
     });
+    
+    models.reset();
+
+    request(process.env.API_URL + "/data/leaderboardRating?include=player&sort=-rating&filter=leaderboard.id==4;updateTime=ge=" +
+      pastMonth.format("YYYY-MM-DDTHH:mm:ss") + "Z", function (error, response, body) {
+      if (error || response.statusCode > 210) {
```
louvenarde, line 100: is that the traditional NPM request we're using here? It's heavily deprecated if so, we should probably not use that anymore (2020)
probably not in the scope of that PR but good to keep in mind

scripts/getAllClans.js
```diff
@@ -1,7 +1,8 @@
-require('dotenv').config();
+require("dotenv").config();
 
-let request = require('request');
-let fs = require('fs');
+let request = require("request");
+let fs = require("fs");
+let moment = require("moment");
```
louvenarde, line 5: all require() => const

scripts/extractor.js
```diff
@@ -2,7 +2,8 @@ require("dotenv").config();
 
 let jsonapi = require("json-api-models");
 let request = require("request");
-let fs = require('fs');
+let fs = require("fs");
+let moment = require("moment");
```
louvenarde, line 6: all require() => const 🙂

louvenarde (commented):
❌  should not have been merged in this state ! ☹

PR #389, opened 2022-06-13
Leaderboards

JaviTrek:


MrRowey (approved):

MrRowey:
@Rackover can you check the JS files and approve them if they are fine :)

README.md
```diff
@@ -43,3 +43,40 @@ As of March 2022 the main 4 Languages that are set up on POEditor are:
...
+
+### Mixins and Variables
+
+Currently, we use the _mixins.sass and _variables.sass to store these elements. Please, try using the already stablished colors and see if a previous mixin by serve your need. The previous website makers did not think about this and so they left 900 lines of variables... Re-using variables will not only do more with less code, but allow us for an easier change in color or adding accesibility buttons (since if we only need to change the color of 3-5 variables, its much easier to install color changing features than lets say 20-30 variables).
+
+If you believe a variable or mixin could be written better, feel free to play with them. Just make sure your changes work well/don't break responsiveness.
+
+### Colors and Aesthetics
+
+Currently, we want the FAF website to have a very gray, black and mute color scheme. With our "highlight" color being the "gold" seen in the buttons. Remember what FAF stands for: Big robots, war, strategy, total control, community, muteness/unsaturated colors, etc. Ergo, try avoiding using flashy colors.
```
louvenarde: This is not a code guideline, and pertains to your personal opinion. it should not drive nor hinder the development of the FAF website, and it does not belong here

README.md
```diff
@@ -43,3 +43,40 @@ As of March 2022 the main 4 Languages that are set up on POEditor are:
...
+
+If you believe a variable or mixin could be written better, feel free to play with them. Just make sure your changes work well/don't break responsiveness.
+
+### Colors and Aesthetics
+
+Currently, we want the FAF website to have a very gray, black and mute color scheme. With our "highlight" color being the "gold" seen in the buttons. Remember what FAF stands for: Big robots, war, strategy, total control, community, muteness/unsaturated colors, etc. Ergo, try avoiding using flashy colors. 
+
+### Artistic Freedom
+
+Don't feel scared or worry about making changes. The FAF website needs as much help as it can get.
```
louvenarde: Not a code guideline either!
If it's open source I think it's a pretty clear message that everyone can make changes :v

public/js/app/index.js
```diff
@@ -1,15 +1,15 @@
...
+ /*async function playerCounterJSON() {
     const response = await fetch('js/app/members/recent.json');
     const data = await response.json();
     return data;
   }
   playerCounterJSON()
     .then((data) => {
       document.getElementById('playerCounter').insertAdjacentHTML("afterbegin", data.length);
-    });
+    }); */
```
louvenarde, line 12: Dead code => Remove
Codebase is not a museum, if you want to look at functions that used to be here you have git 😄

public/js/app/index.js
```diff
@@ -112,7 +112,7 @@ document.addEventListener("DOMContentLoaded", () => {
 
   window.addEventListener('scroll', () => {
     handleScrollAnimation();
-    console.log("You scrolled!");
+    //console.log("You scrolled!");
```
louvenarde, line 115: Remove it

public/js/app/leaderboards.js
```diff
@@ -1,48 +1,36 @@
...
-      
-    }
-  }
-  //console.log(`We are missing ${missingPlayers} players`);
-  let dataMinusMissingPlayer = (data.length - missingPlayers).toString();
-
-  if (dataMinusMissingPlayer.length !== 4) {
-    lastPage = dataMinusMissingPlayer.slice(0, 1);
+function leaderboardUpdate() {
+  // We convert playerList into a string to find out how many pages we have. We can find so by checking the first two digits. In other words, if we have 1349 players, then we have 13 pages. However, if we have 834, we have 8 pages (only take the first digit).
```
louvenarde, line 27: remove all of this and just divide by 100, then floor. This is how you get the "first two digits" of a number... 😅

louvenarde: 2468 
/100 = 24.68
Floored => 24

public/js/app/leaderboards.js
```diff
@@ -51,185 +39,130 @@ function leaderboardUpdate(data) {
...
-    //console.log(`data.length minus missing players is ${data.length - missingPlayers}`);
-    //console.log(` for loop data.length is ${data.length - 100 - (pageNumber * 100)}`)
-    
 
     if (playerIndex <= 0) {
-      //console.log('I broke');
-      playerIndex -= (playerIndex) * -1;
-    } else {
-      console.log(` We didn't break! `);
+      console.log('I broke');
```
louvenarde: "I broke" is not a valid error message. Try specifying exactly what went wrong, how to fix it and what to look for in the callstack. think of future developers who will have to debug your thing.

public/js/app/leaderboards.js
```diff
@@ -51,185 +39,130 @@ function leaderboardUpdate(data) {
...
     //Switches background color for even and uneven
-    if (currentDate >= data[playerIndex].date) {
-      next100Players--;
-      addNextPlayer++;
-      previousMissingPlayer++;
-      //console.log(`${data[playerIndex].label} didn't play recently`); 
-      
-      
+    if (backgroundColor === 'greyLeaderboard') {
+      backgroundColor = 'darkGreyLeaderboard';
```
louvenarde: 1) Why ?
2) Use constants, no magics

public/js/app/leaderboards.js
```diff
@@ -51,185 +39,130 @@ function leaderboardUpdate(data) {
...
+      backgroundColor = 'greyLeaderboard';
     }
+    // Gets the player data and inserts it into the li element
+    let rating = playerList[playerIndex].value.rating;
+    let winRate = playerList[playerIndex].value.wonGames / playerList[playerIndex].value.totalgames * 100;
+    document.getElementById('onePlayer').insertAdjacentHTML('afterbegin', `<li class='leaderboardDelete ${backgroundColor}' > ${playerList[playerIndex].label}</li>`);
+    document.getElementById('oneRank').insertAdjacentHTML('afterbegin', `<li class='leaderboardDelete ${backgroundColor}'> ${-1 * (playerIndex - playerList.length)}</li>`);
+    document.getElementById('oneRating').insertAdjacentHTML('afterbegin', `<li class='leaderboardDelete ${backgroundColor}'> ${rating.toFixed(0)}</li>`);
+    document.getElementById('oneGamesAmount').insertAdjacentHTML('afterbegin', `<li class='leaderboardDelete ${backgroundColor}'> ${playerList[playerIndex].value.totalgames}</li>`);
+    document.getElementById('oneWon').insertAdjacentHTML('afterbegin', `<li class='leaderboardDelete ${backgroundColor}'> ${winRate.toFixed(1)}% </li>`);
```
louvenarde: Instead of insertHTML you need to make a helper function that creates the elements (li) and then sets it as child of the DOM of your onerating/onerank/onewon thingy. Modifying HTML at runtime refreshes DOM and lags a bunch, plus it's not very clean! You also have a lot of repetition here, you can factorize that in a function

public/js/app/leaderboards.js
```diff
@@ -51,185 +39,130 @@ function leaderboardUpdate(data) {
...
+let pageButton = document.querySelectorAll('.pageButton');
+function pageChange(newPageNumber) {
+  pageNumber = newPageNumber;
+  pageButton.forEach(element => element.classList.remove(`exhaustedButton`));
+  if (pageNumber === 0) {
+    //You see 4-7 pageButton because there are a total of 8 buttons counting the ones at the bottom of the page
+    pageButton[0].classList.add('exhaustedButton');
+    pageButton[3].classList.add('exhaustedButton');
+    pageButton[4].classList.add('exhaustedButton');
+    pageButton[7].classList.add('exhaustedButton');
```
louvenarde, line 65: If someone modifies the HTML code or the order these buttons are displayed, your code breaks entirely without warning.
If these are different buttons, give them each a specific ID and disable them specifically
If these are not different buttons, make a for loop to group buttons and iterate over them to disable them in a single statement!

public/js/app/leaderboards.js
```diff
@@ -51,185 +39,130 @@ function leaderboardUpdate(data) {
...
     .then(data => {
-      leaderboardUpdate(data);
+      playerList = data;
+      let dataLengthString = (playerList.length).toString();
+      if (dataLengthString.length !== 4) {
+        lastPage = dataLengthString.slice(0, 1);
+      } else {
+        lastPage = dataLengthString.slice(0, 2);
+      }
+      console.log(playerList.length);
```
louvenarde: Remove log

public/js/app/leaderboards.js
```diff
@@ -51,185 +39,130 @@ function leaderboardUpdate(data) {
...
-function changePage() {
-  leaderboardOneJSON(currentLeaderboard)
+// This changes the current leaderboard(newLeaderboard), sets the page to 0 (pageNumber = 0) and resets the next and previous buttons.
+function changeLeaderboard(newLeaderboard) {
+  leaderboardOneJSON(newLeaderboard)
     .then(data => {
-      leaderboardUpdate(data);
+      playerList = data;
+      let dataLengthString = (playerList.length).toString();
+      if (dataLengthString.length !== 4) {
```
louvenarde: Just divide by 100 and floor! do not convert to string / splice 🙂

public/js/app/leaderboards.js
```diff
@@ -51,185 +39,130 @@ function leaderboardUpdate(data) {
...
+        lastPage = dataLengthString.slice(0, 1);
+      } else {
+        lastPage = dataLengthString.slice(0, 2);
+      }
+      console.log(playerList.length);
+      pageChange(0);
     });
 }
+//Gets called once so it generates a leaderboard
+changeLeaderboard('1v1');
```
louvenarde, line 114: Use constant, no magic declaration in code
const defaultLeaderboardName = '1v1' in the beginning of your script

public/styles/site/leaderboards.sass
```diff
@@ -25,22 +26,33 @@ input
...
+  text-align: center
+  option
+    color: variables.$color-active
+    background-color: variables.$background-secondary
+      
+// This is in here for the js since the leaderboard is actually starting from 100 to 1. So it needs to be displayed in reverse.
+.rotateMe180
+  ul
+    li
+      transform: rotate(180deg)
```
louvenarde: ouch! Is that the best thing we can do? Transform rotate is quite expensive
You can reverse order elements with a flexbox for instance, but there are other ways to do it in CSS
but do we even need to do it in CSS? Can't we just reverse the JS array before displaying...?

scripts/extractor.js
```diff
@@ -48,10 +48,10 @@ module.exports.run = function run() {
 
     const models = new jsonapi.Store();
 
-    var pastMonth = moment().subtract(3, 'months');
+    var pastMonth = moment().subtract(6, 'months');
 
     request(process.env.API_URL + "/data/leaderboardRating?include=player&sort=-rating&filter=leaderboard.id==1;updateTime=ge=" +
-      pastMonth.format("YYYY-MM-DDTHH:mm:ss") + "Z" + '&page[size]=5000', function (error, response, body) {
+      pastMonth.format("YYYY-MM-DDTHH:mm:ss") + "Z" + '&page[size]=9499', function (error, response, body) {
```
louvenarde: 9499 ? why ?

scripts/extractor.js
```diff
@@ -67,7 +67,7 @@ module.exports.run = function run() {
     models.reset();
 
     request(process.env.API_URL + "/data/leaderboardRating?include=player&sort=-rating&filter=leaderboard.id==2;updateTime=ge=" +
-      pastMonth.format("YYYY-MM-DDTHH:mm:ss") + "Z" + '&page[size]=5000', function (error, response, body) {
+      pastMonth.format("YYYY-MM-DDTHH:mm:ss") + "Z" + '&page[size]=9499', function (error, response, body) {
```
louvenarde: Same

scripts/extractor.js
```diff
@@ -83,7 +83,7 @@ module.exports.run = function run() {
     models.reset();
 
     request(process.env.API_URL + "/data/leaderboardRating?include=player&sort=-rating&filter=leaderboard.id==3;updateTime=ge=" +
-      pastMonth.format("YYYY-MM-DDTHH:mm:ss") + "Z" + '&page[size]=5000', function (error, response, body) {
+      pastMonth.format("YYYY-MM-DDTHH:mm:ss") + "Z" + '&page[size]=9499', function (error, response, body) {
```
louvenarde: Same

scripts/extractor.js
```diff
@@ -99,7 +99,7 @@ module.exports.run = function run() {
     models.reset();
 
     request(process.env.API_URL + "/data/leaderboardRating?include=player&sort=-rating&filter=leaderboard.id==4;updateTime=ge=" +
-      pastMonth.format("YYYY-MM-DDTHH:mm:ss") + "Z" + '&page[size]=5000', function (error, response, body) {
+      pastMonth.format("YYYY-MM-DDTHH:mm:ss") + "Z" + '&page[size]=9499', function (error, response, body) {
```
louvenarde: Same

scripts/extractor.js
```diff
@@ -83,7 +83,7 @@ module.exports.run = function run() {
     models.reset();
 
     request(process.env.API_URL + "/data/leaderboardRating?include=player&sort=-rating&filter=leaderboard.id==3;updateTime=ge=" +
```
louvenarde: This is one hell of a lot of code duplication for something that gets so repeated, you need to make a For loop, fetch every leaderboard from the API like the client does and then do your request once per leaderboard using their IDs you fetched. This hardcode won't do and it's a hazard for the day we modify leaderboards

templates/views/leaderboards.pug
```diff
@@ -46,50 +43,62 @@ block content
...
+        ul
+          li(onclick= `pageChange(pageNumber + 1)`).pageButton Next
+          li(onclick= `pageChange(pageNumber - 1)`).pageButton.exhaustedButton Previous 
 
           
     
 block js
 
-  script( src="../../js/app/leaderboards.js")
+  script( src="../../js/app/leaderboards.js" )
```
louvenarde: huh ? ` ` ?

louvenarde (changes requested):
Needs quite a bit of work I'm afraid! Now this is your own fork so you don't have to address my review for this if you don't want to
@MrRowey I can help with code & review but if you guys want to merge stuff on your own branch i'm not stopping you. However I will require reviews for merging anything into develop 😄

public/js/app/leaderboards.js
JaviTrek: but I need the background color to change for even and uneven players. If I make it a constant, wouldnt it stay constantly in one color?

louvenarde: Use even and odd CSS rules ! https://www.w3.org/Style/Examples/007/evenodd.en.html 🙂
No need for javascript for that

public/js/app/leaderboards.js
JaviTrek: `
function createLi(playerNumber) {

  let createLi = document.createElement('li');
  createLi.classList.add('leaderboardDelete', `${backgroundColor}`);
  let textPlayerName = document.createTextNode(`${playerList[playerNumber].label}`);
  createLi.appendChild(textPlayerName);
  let leaderboardPlayer = document.getElementById(`leaderboardPlayer`);
  leaderboardPlayer.appendChild(createLi);


}
`

Are you sure this is a better idea? I would need this about 10 times, adding an obscene amount of code to the js file. Seems like a very complicated way of dealing with such a simple issue. Not sure if there is a better way to do this since I haven't touched child and these sort of things ever

louvenarde: why would you need this 10 times, does something prevent you from passing the changing parameters as part of the function arguments? :v

public/js/app/leaderboards.js
JaviTrek: This was very helpful, thanks!

public/js/app/leaderboards.js
JaviTrek: Oh this was very useful as well, thanks alot!

public/js/app/leaderboards.js
JaviTrek: But then I would need to call the function 10 times if I need to change the parameters.

louvenarde: Which is okay, 10 function calls will take 10 lines at most, not really an "obscene amount of code" :p 
You can do it all in a function called "RefreshLeaderboardPages" or whatever, it's fine

JaviTrek: well yeah, I meant doing 10 times the components of the function but calling it many times would work aswell. Also do you know where could I find more about how the insertadjacent html lags a bunch? I couldnt find much but then again, I'm not very knowledgable about this stuff

louvenarde: You can look for performance comparisons online and you'll see that it's slower because it requires parsing. From experience I know modifying / injecting HTML code directly in the page can refresh an undisclosed amount of DOM elements at once, which in the past has caused me extreme lag in some cases. It's just worse than modifying the DOM tree specifically with precises operations, so as it takes the same amount of space to factorize it & since it's also better practice I would enjoin you to do it. But it's your code on your branch, you do you 😁  As long as this never gets into develop it's fine

louvenarde: Here's a page I found by googling a bit https://medium.com/codex/what-is-faster-to-insert-into-dom-html-or-dom-nodes-ff11586f8570 but there's a lot of benchmrking on webjs stuff online so feel free to look for other comparisons!

PR #403, opened 2022-08-01
Client News Bug Fix

JaviTrek:
## The Bug
After the News section has at least been opened once, then switched to another tab and then come back to News, all links become internal (meaning the links are being opened inside the client rather than outside in a browser)

### Solution

The JS has a line at the end to also set all links' (a) target to _blank.

### Other
Now the newshub also updates every 5 minutes rather than 15.


MrRowey:
@Sheikah45 this needed a ENV change for the update interval which repo dose this need a extra PR for adding this ?

Sheikah45:
You can add it to the example in faf stack but really you need to remind me to set the variable when deploying

PR #405, opened 2022-09-14
Added a fix for the bad image downscaling

phongu:
Attempt at fixing the low quality image downscaling on the newshub using some canvas elements that sit inside the image divs and a purpose-built library: https://github.com/nodeca/pica

### DIFF
--- a/public/js/app/newshub.js
+++ b/public/js/app/newshub.js
@@ -15,6 +15,66 @@ let clientSpawn = document.getElementById('clientSpawn');
 let clientContainer = document.querySelectorAll('.clientContainer');
 let clientMainFeature = document.querySelectorAll('.clientMainFeature');
 
+//downscale-fix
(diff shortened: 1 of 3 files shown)


phongu:
Fixed the CORS issue, with help from Brutus

JaviTrek (commented):
The changes seem to not affect the current newshub in any negative way. Have you tested this changes in a local version of the FAF client @phongu ? If so, I'll Approve it/give the approving review. Thanks for your work!

phongu:
Javi yes i tested on the faf client and it works.

JaviTrek (approved):
I have just tested this locally inside the FAF client. All seems to be working correctly.
image


### On a separate note
Approving these changes will also bring my PR from like a month (or two?) ago where I fixed the client sometimes opening links inside the client instead of a new tab in the browser.

PR #406, opened 2022-09-28
Update code of conduct

BlackYps:
The discussion happened in the board a while ago, but then I forgot to have it actually published.

The name also changed from Code of Conduct to Contribution guidelines, but I don't know what else needs to change if I would change the file name


MrRowey:
yeh theirs a few place that it need to change if were to change the file name the easy thing to do for now would be to just change the title for them in the https://github.com/FAForever/website/blob/develop/templates/layouts/default.pug default layout .

@JaviTrek Make sure this is replicated to the new-frontend branch

MrRowey (approved):

MrRowey:
If your renaming those files then you need to also change in `express.js` line 97

BlackYps:
Oh, you originally meant only changing the title, not the filename xD

PR #411, opened 2023-01-02
Error Handling

JaviTrek:
Small Fixes regarding the error handling the website has in-case the wordpress or API are down so the whole website doesn't crash.


.env.example
```diff
@@ -1,16 +1,24 @@
...
 OAUTH_CLIENT_SECRET=banana
 SESSION_SECRET_KEY=banana
 RECAPTCHA_SITE_KEY=6Lc4GmgaAAAAAPaFbnwuP_dBGG--SSy2Hi5UcqwK
 
+
 TOKEN_LIFESPAN=43200                       # an api token is valid for 12h
+LEADERBOARDS_UPDATE_INTERVAL=900          # 15 minutes
+RECENT_USERS_LIST_UPDATE_INTERVAL=900     # 15 minutes
+CLIENT_RELEASE_FETCHING_INTERVAL=900      # 15 minutes
+
```
Sheikah45: Are these supposed to be back?

public/js/app/getClans.js
```diff
@@ -19,8 +19,8 @@ async function getClan() {
 
   // We compare the url TAG with the TAGS available in getAllClans and find the clan leader this way
   
-  
-  const response = await fetch(`https://api.test.faforever.com/data/clan?include=memberships.player&filter=tag==${clanTag}`);
+  // TODO: Change this hardcoded url into something with env
```
Sheikah45, line 22: we already have process.env.API_URL so should use that here now

JaviTrek: I don't think we can access env variables from the front end. This needs to be re-done with the backend rather than the frontend (which is now possible with my local fullstack enviroment thank god)

.env.example
JaviTrek: No, I'm not sure why they propped back, probably an error of mine while resolving the conflicts...

public/js/app/getClans.js
Sheikah45: ah got it didn't realize this was frontend my bad

Sheikah45 (approved):

PR #415, opened 2023-01-02
Flash Message Fix and client newshub

JaviTrek:
Fix done in order to pass the faf test. Apparently the flash messages can break easily. Small client newshub fix so it doesn't open browser links inside the client.

### DIFF
--- a/templates/mixins/flash-messages.pug
+++ b/templates/mixins/flash-messages.pug
@@ -3,5 +3,6 @@ mixin flash-messages(messages)
     div.alert(class=flash['class'])
       ul.flash-errors
         if flash.messages
-          if flash.messages[0].msg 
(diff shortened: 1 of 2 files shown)


templates/mixins/flash-messages.pug
```diff
@@ -3,7 +3,5 @@ mixin flash-messages(messages)
     div.alert(class=flash['class'])
       ul.flash-errors
         if flash.messages
-          if flash.messages[0].msg
+          if flash.messages[0].msg 
             li <strong>#{flash.type}</strong> !{flash.messages[0].msg}
```
Sheikah45: I would still add the if flash.messages[0]
to protect when the object is undefined

Sheikah45 (changes requested):

Sheikah45 (approved):

PR #417, opened 2023-01-03
Add catch to getRepositoryPromise

Sheikah45:


JaviTrek (approved):

Sheikah45:
This doesn't work because the promise is never rejected with a 403 error from the getRelease as per https://github.com/github-tools/github/issues/439

PR #428, opened 2023-01-06
Steam Link, Navigation, lobby_api

JaviTrek:
### SteamLink
Steamlinking routes have been all transformed back to their outdated routes for the sake of a hotfix in order for steamlink to work properly.

### Navigation
Received small superficial update to be better looking and give users a better experience

### lobby_api
reinstated after it seems it is actually being used and is not some useless piece of code. It provides users with a mirror of the users, their maps and lobbies in the FAF client.


express.js
```diff
@@ -238,8 +243,10 @@ passport.use('faforever', new OidcStrategy({
             return verified(null);
           }
           let user = JSON.parse(body);
+          console.log(accessToken)
           user.data.attributes.token = accessToken;
           user.data.id = user.data.attributes.userId;
+          console.log(user.data.attributes);
```
Sheikah45, line 249: These logs should be removed

express.js
```diff
@@ -254,7 +261,9 @@ passport.use('faforever', new OidcStrategy({
 
 
 passport.serializeUser(function (user, done) {
+  console.log(user.data.attributes);
```
Sheikah45, line 264: Same here

Sheikah45 (changes requested):

PR #434, opened 2023-01-15
Backend Update

JaviTrek:
New Update for multiple things relating to the backend.

### Axios in, Request out
Alot of old routes are being updated to use less code and modern dependencies such as Axios. We currently use Request which is deprecated and not maintained since 2020 and Axios provides better error handling and more up to date stuff.

### We dont need 3 pages to change email, username and password.
All these have been summed up to one page called (account) settings where the user can change all 3. Other features should be added here as well such as clan management and so on.

### getClan/seeClan works optimally
before, when we would see a specific clan, the page would load and wait about 3 seconds before loading the clan data. Now the page doesn't open until the clan info is loaded. Plus the API call now is done through the env API_URL rather than a hardcoded API in the frontend (which can't use env variables, only backend stuff can do it). It also is much easier for people to understand the url (showing tag of the clan in url such as ?tag=ANZ rather than ?id=1234) so the username can just quickly switch the tag if needed.


JaviTrek:
@Sheikah45 I've tested everything back and forth and most changes weren't significant/destructive so I believe everything should work well but if you want to review this since its a bit bigger update, feel free. Otherwise, I personally believe its rready for production/wont cause any big problems. Thanks for your time!

routes/lobby_api.js
```diff
@@ -2,12 +2,12 @@ const request = require("request");
...
 let isFetching = false;
 
-const PLAYER_COUNT_UPDATE_INTERVAL = parseInt(process.env.PLAYER_COUNT_UPDATE_INTERVAL) * 1000;
+const PLAYER_COUNT_INTERVAL = process.env.PLAYER_COUNT_INTERVAL * 1000;
 
 exports = module.exports = function (req, res) {
   let resource = req.query.resource;
   if (cache[resource]) {
-    if (isFetching || Date.now() - cache[resource].pollTime < PLAYER_COUNT_UPDATE_INTERVAL) {
+    if (isFetching || Date.now() - cache[resource].pollTime < PLAYER_COUNT_INTERVAL) {
```
Sheikah45, line 11: Why change this variable name?

Sheikah45 (changes requested):
I have this live on the test server and can't open the menu when I click on my name after logging in

Edit: nevermind might have just been a transient error

Sheikah45:
Also the reset password link at the top disappeared and that needs to exist without login info

Sheikah45:
Also all the direct links to the settings like account/changeEmail still just open the change username page. They should at least open the proper window of the page

routes/lobby_api.js
JaviTrek: Because PLAYER_COUNT_UPDATE_INTERVAL didn't exist under .env.example so I just set it to something else with a short time limit (I believe 15 seconds?).

JaviTrek:
> Also the reset password link at the top disappeared and that needs to exist without login info

The reset password can be accessed through the login page, it has a "forgot your password" link that will take you to it. This is the modern standard everywhere afaik, all sites have a forgot your password in the login area. Having an additional link for it is unnecessary.

JaviTrek:
> Also all the direct links to the settings like account/changeEmail still just open the change username page. They should at least open the proper window of the page

Just added/fixed this! However, it just made me realize, the password changing page is a bit broken, the new passwords dont need to match to be changed, you just need the current password and then type a new password without checking if the new password was typed twice correctly.

JaviTrek:
I've been a bit busy with real life and uni but this is almost done. It just needs a touch of polishing and testing to make sure everything is working correctly. I haven't abandoned this, simply put it in pause to focus on other things.

express.js
```diff
@@ -275,9 +279,9 @@ app.get(`/${process.env.CALLBACK}`, passport.authenticate('faforever', {
   failureRedirect: '/login', // Failed auth
   failureFlash: true
 }), function (req, res) {
-  res.redirect('/');
-  //res.redirect(fullUrl ? fullUrl : '/');
-  //fullUrl = '/'; // Successful auth
+  //Success Auth
+  res.redirect(previousURL ? previousURL : '/');
```
fcaps, line 283: this will not work with multiple users, previousURL is a global variable (used by all users)

.env.example
```diff
@@ -4,14 +4,15 @@ HOST=http://localhost:3000
...
+LOBBY_API_URL=http://lobby.faforever.com:4000
 CALLBACK=auth
 OAUTH_CLIENT_ID=faf-website
 OAUTH_CLIENT_SECRET=banana
 SESSION_SECRET_KEY=banana
 RECAPTCHA_SITE_KEY=6Lc4GmgaAAAAAPaFbnwuP_dBGG--SSy2Hi5UcqwK
 
 
-TOKEN_LIFESPAN=43200                       # an api token is valid for 12h
+TOKEN_LIFESPAN= 12                       # an api token is valid for 12h
```
fcaps, line 15: you cannot just change a env file variable meaning.. everyone who had this already installed will not get the change and the website will just do weird stuff.

fcaps:
This PR is outdated.
Currently scraping some stuff out of it, but there is no hope left getting this PR merged without multiple conflicts and creating issues.

fcaps:
@JaviTrek this is finally outdated, you can cherrypick the changes on account and create a new PR, but i'll close this one.

PR #450, opened 2023-11-09
#449 remove npm lock

### LINKED ISSUE #449: [cleanup] remove npm lock
we using yarn in the dockerfile, but a package-lock.json is still present.
non-critical change , since it will not have an effect on the docker-image.

Output from a docker-build:
`warning package-lock.json found. Your project contains lock files generated by tools other than Yarn. It is advised not to mix package managers in order to avoid resolution inconsistencies caused by unsynchronized lock files. To clear this warning, remove package-lock.json.`

how a package should be added:
`yarn add [package-name] [--dev]`

examples:
for production packages 
`yarn add my-package `

for dev only:
`yarn add my-package --dev`

Old and not supported:
`npm install my-package`

fcaps:
closes #449 

non critical, since this is ignored anyways in the build.
dependabot should also


fcaps:
dependabot is also fixed with that (was picking up updates of packages, that are not used in the image)

Brutus5000:
Shouldn't we do it the other way round and make sure the build process considers the lock file?

fcaps:
already a fix for this on the way.

fcaps:
we are using yarn, so only the "yarn.lock" is used. the package-lock is unused (maybe some dev was executing npm install instead of yarn install). 
since there is no check in the Dockerfile with `--frozen-lockfile`,  the yarn installation was only complaining about it with: 

`"warning package-lock.json found. Your project contains lock files generated by tools other than Yarn. It is advised not to mix package managers in order to avoid resolution inconsistencies caused by unsynchronized lock files. To clear this warning, remove package-lock.json."`

Sheikah45:
This is no longer required correct?

fcaps:
correct

PR #410, opened 2022-12-29
New frontend

JaviTrek:
# The new FAF website is out!
After one year, the new website is finally coming out to production. Below you will find the most significant changes done.
image

# Backend
Overall the backend hasn't received many significant changes. It is the next step for the website to uninstall the request package with axios and clean up alot of the backend routes that repeat the same code (could be solved with a middleware maybe?). Nonetheless, the new backend is much more cleaner and easier to read since a lot of fat was cut.
## Routes
Routes in the express.js have received major changes. They've all received the adequate loops needed to look cleaner (rather than repeating the same route.get for each individual page in the website). Therefore making it much cleaner and easier to read. 
Routes have also been simplified. Previously, each page had its own route js page for its own custom middleware. However, alot of pages didn't need any middleware but simply display information in the frontend. Thanks to this simplification, about 21 pages in routes have been deleted.

## News articles
On the /scripts/getNews.js script, we now have the capability to take blog posts from the wordpress and display them as complete news articles (long gone are the days of needing to display everything long in the forums). The getNews.js gets the news from the wordpress and creates routes/pages for each article found (meaning, we don't need to make manual commits for each news article URL or page)
image
image

## API calls/extractor.js
Also on /scripts/extractor.js there's been improvement. Alot of API calls have been in a compact for loop in order to reduce code repetition. These API calls get cached in json files and reduce the amount of times the website pings the API.

## Renaming/Re-pathing
A couple of the routes have received minor name changes in order to illustrate their function better (such as link.js becoming linkSteam.js) and pathing made much more simpler (for example, account/password/reset is now account/changePassword which is more simple and password doesn't need its own path).

# Frontend
The frontend has received the bigger batch of changes. Ranging from new pages to overhauls in others.

## New aesthetic/theme
The new website now has a theme of primary grey - black with a secondary orange/gold - dark red colors. I believe this suits us better than the white and black theme of the old website.

## Update the old pages
Anything ranging from the home page to the leaderboards, all previous pages (which weren't alot) have received updates to their information and interface. Making it look more modern and easier to use or faster to load for our users.

## New pages
Now we have from contribution, donation, AI, campaign, FAF teams and alot other pages in our website. All of these serve the great purpose of having a more official page to link users to rather than linking them to forum or wiki posts (for example, steam vs FAF changes, introduction guides, changes in the co-op campaign). Alot of these pages are re-using information from said forum and wiki posts (and sometimes even link to them in-case user wants more information). 

##  Play now page
[...]

PR #453, opened 2023-11-10
cleanup, local development fixes and start best practices

fcaps:
Cleanup, local development improvements and applying some best practices
- removed some unused yarn deps (via npm_check)
- added missing deps (grunt did not complain on concat, but files where missing)
- removed Procfile (was prob. a leftover of a Heroku trail)
- removed .dockerignore, since it is doing nothing for the build
- fixed all of the grunt commands (that are used)
- removed grunt node-inspector -> unused
- added npm_check to find unused yarn deps and updates
- Dockerfile was updated to include some of the best practices for node-containers
- Docker compose was added for easy development with the faf-stack
- Readme updated to include local docker env setup
- Express port is now defaulting to 3000
- OAUTH_PUBLIC_URL was intruduced for docker environments (container -> container vs brower -> container) urls
- unused js files removed (no reference in the code, or empty files)


fcaps:
we can split this PR into multiple smaller ones, but keeping all the things in my head (what js file can be deleted etc,) was too much for me.

.dockerignore
fcaps, line 1: ignoring node_modules did nothing for the build nor the local env

.env.faf-stack
```diff
@@ -0,0 +1,33 @@
...
+LOBBY_API_URL=http://faf-python-server:4000
+
+OAUTH_CLIENT_ID=faf-website
+OAUTH_CLIENT_SECRET=banana
+SESSION_SECRET_KEY=banana
+RECAPTCHA_SITE_KEY=6LeIxAcTAAAAAJcZVRqyHh71UMIEGNQ_MXjiZKhI
+TOKEN_LIFESPAN=43200
+CLAN_INVITES_LIFESPAN_DAYS=30
+EXTRACTOR_INTERVAL=5
+PLAYER_COUNT_INTERVAL=15
```
fcaps, line 36: we should also update the faf-stack to include some of the fixes from here.

.env.example
```diff
@@ -10,11 +9,7 @@ OAUTH_CLIENT_ID=faf-website
...
-TOKEN_LIFESPAN=43200                       # an api token is valid for 12h
-CLAN_INVITES_LIFESPAN_DAYS = 30            # 30 days
-EXTRACTOR_INTERVAL= 5 # 5 minutes
-PLAYER_COUNT_INTERVAL= 15           # 15 seconds
-
-
+TOKEN_LIFESPAN=43200
+CLAN_INVITES_LIFESPAN_DAYS=30
+EXTRACTOR_INTERVAL=5
+PLAYER_COUNT_INTERVAL=15
```
Sheikah45: Why were the comments removed, in this case at least they help inform about units where the names are insufficient for that

.env.faf-stack
```diff
@@ -0,0 +1,33 @@
+# node env is set statically to "development" in the dockerfile, you could override it here.
```
Sheikah45, line 1: What was the thinking with having both a .env.template and .env.fafstack? It seems a little redundant

Dockerfile
```diff
@@ -1,27 +1,22 @@
...
+COPY --from=builder --chown=node:node /code /code
 
 WORKDIR /code
+USER node
 
-# Only install runtime dependencies for the runtime image
-RUN yarn --prod
-
-CMD PORT=3000 npm start
+CMD ["dumb-init", "node", "express.js"]
```
fcaps, line 21: applied some of the best practices here.
more info https://snyk.io/blog/10-best-practices-to-containerize-nodejs-web-applications-with-docker/

Dockerfile-dev
```diff
@@ -0,0 +1,6 @@
+FROM node:20.9-bookworm
+RUN apt-get update && apt-get install -y --no-install-recommends dumb-init
+ENV NODE_ENV development
+WORKDIR /code
+
+CMD ["dumb-init", "./node_modules/.bin/grunt", "concurrent"]
```
fcaps, line 6: multistage was actually not beneficial, so i decided to create a new dockerfile for local-dev

Procfile
fcaps, line 1: we are not using heroku somehow, right?

docker-compose.yml
```diff
@@ -0,0 +1,19 @@
...
+      - .:/code
+    build:
+      dockerfile: Dockerfile-dev
+      context: .
+    networks:
+      -  faf-stack
+    
+networks:
+  faf-stack:
+    name: faf-stack_faf
```
Sheikah45, line 18: I wouldn't call it faf_stack since realistically that is dying soon TM

express.js
```diff
@@ -17,9 +17,10 @@ app.locals.clanInvitations = {};
 require('dotenv').config();
 //Define environment variables with default values
 process.env.NODE_ENV = process.env.NODE_ENV || 'production';
-process.env.PORT = process.env.PORT || '4000';
+process.env.PORT = process.env.PORT || '3000';
```
fcaps, line 20: in every .env file i could found we setting the `PORT ` env, just because we have this fallback here.

express.js
```diff
@@ -227,7 +228,7 @@ app.get('/login', passport.authenticate('faforever'));
 passport.use('faforever', new OidcStrategy({
     issuer: process.env.OAUTH_URL + '/',
     tokenURL: process.env.OAUTH_URL + '/oauth2/token',
-    authorizationURL: process.env.OAUTH_URL + '/oauth2/auth',
+    authorizationURL: process.env.OAUTH_PUBLIC_URL + '/oauth2/auth',
```
fcaps, line 231: a browser can not resolve the docker name aka "http://faf-ory-hydra:4444", so this is needed for docker environments, this needs also be fixed in the faf-stack

Sheikah45 (commented):
I am not as familiar with grunt so can't comment too much on that

grunt/concat.js
fcaps, line 1: none of the removed files where used, sometimes they where just empty after  #410

grunt/concurrent.js
```diff
@@ -1,6 +1,6 @@
 module.exports = {
 	dev: {
-		tasks: ['nodemon:debug', 'concat:js', 'uglify:dev', 'watch'],
+		tasks: ['nodemon', 'concat', 'watch'],
```
fcaps, line 3: uglify is not needed on dev, and calling a taks without `:`, just executes all tasks

docker-compose.yml
fcaps: docker needs it like this, otherwise i cannot access the containers from the faf-stack.
more info: https://stackoverflow.com/questions/38088279/communication-between-multiple-docker-compose-projects

.env.faf-stack
fcaps: env.example is for local development (without docker and a faf-stack)

.env.example
fcaps: was removing the spaces (spaces are not save in env files), but then i thought the comments where kinda useless - at least for me. i can add them again if you liked them

Sheikah45: I would keep the comments since they indicate the units

docker-compose.yml
Sheikah45: got it didn't realize this was supposed to be used in conjunction with faf_stack

.env.example
fcaps: added them again.

Sheikah45 (approved):

Issue #445, opened 2023-11-04
Multiple features (clans, leaderboard etc.) not working.


fcaps:
Due some changes (Auth required for the DataController) in the[ java-api](https://github.com/FAForever/faf-java-api/commit/3a45fe741b7053605df6f415481a03f9790f6ddb#diff-4755b3b8645a7df33256e5fa3751d71ad176f174f3f1dfc5f31edabae05ba976L38), the leaderboard, clans and other portions are not working anymore. 

Issues:
1. the website calls the java-api to "cache" some of the results  [here](https://github.com/FAForever/website/blob/develop/scripts/extractor.js#L202), so we need the client-credentials flow added to the existing hydra-client. -> maybe even create a new client in hydra
2. there is already a [PR ](https://github.com/FAForever/website/pull/434) refactoring some portions of the code that would simplify the bugfix. (moved api-calls to the node backed - currently some are called directly in the browser 

i could work on a bugfix, but don't know if i should fix/finish the other PR, creating a bugfix for that what is worth fixing, or something else bc. you guys already decided to remove it completely.

Since nobody **really** complained about this issue in 2 weeks, i wonder how often the clan management is used.

Brutus5000:
It's even worse than that. Since this app has almost no exception handling, so clan calls make it hard crash.
There is already a OAuth client in Hydra since the API handles registration, password change etc.

Feel free to continue on the existing PR. I don't think anybody minds you finishing it ;)
By the way, I don't recognize your username, please PM me in Discord for Zulip invite if interested.

JaviTrek:
@fcaps I'd recommend finishing my PR, afaik, most of the work should be done but I was too lazy to test out everything. I might finish it soon myself since it breaks my heart seeing the website broken. 

Let me know if you do start on it because I want to make sure we don't work on the same thing!

fcaps:
TL:DR 
Someone has to decide on a solution before we can start, i tried my best to show the issues/implications here.

@JaviTrek would be nice if we could break down tasks of the ongoing PR, so working on this together is possible.
As a start splitting the branch in account and clan stuff would be great, so we don't have a big PR with multiple features going on.

When we have a decision about the solution, we could figure it out what base-branch to use (develop or account-management). Creating merge conflicts is surely nothing we wanna have here.

Issue that i analyzed:

### Clan List ([Link](https://faforever.com/clans))
depends on a cache file `public/js/members/allClans.json` that is created/updated via extractor.js to reduce load on the java-api.
[...]

PR #455, opened 2023-11-11
Disable failing routes and cronjobs. part of #445

fcaps:
All routes for clan and leaderboard are disabled with a new 503 view:
The links are still there, but will lead to this page.
image

### DIFF
--- a/express.js
+++ b/express.js
@@ -105,6 +105,10 @@ const appGetRouteArray = [
 
 //Renders every page written above
 appGetRouteArray.forEach(page => app.get(`/${page}`, (req, res) => {
+  // disabled due https://github.com/FAForever/website/issues/445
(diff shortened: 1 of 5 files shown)


Sheikah45:
I think your branch has stray commits. We squash and merge so you have a few commits from the last PR I believe that can be removed.

fcaps:
> I think your branch has stray commits. We squash and merge so you have a few commits from the last PR I believe that can be removed.

updated

Sheikah45:
It still is showing the extraneous commits because you just merged the remote branch in. It will probably be better if you rebase your branch and drop the duplicate commits

fcaps:
yeah, long time ago since i worked with squash commits.
now it's good

Sheikah45 (approved):

PR #458, opened 2023-11-11
minor issues with navigation and error-handling

### LINKED ISSUE #456: Uncaught SyntaxError: Unexpected end of input
If wordpress is not running, the website generates issues on all sites with:
`Uncaught SyntaxError: Unexpected end of input`
this comes from `/public/js/app/members/flashMessage.json` has no data and the code was poorly written without that in mind.

### LINKED ISSUE #457: Uncaught TypeError: Cannot read properties of undefined (reading 'classList')
Hovering over the login button will create errors like
````
Uncaught TypeError: Cannot read properties of null (reading 'addEventListener')
Uncaught TypeError: Cannot read properties of undefined (reading 'classList')
````

this is because some of html elements are only present if logged in.

fcaps:
closes #456 
closes #457

### DIFF
--- a/public/js/app/navigation.js
+++ b/public/js/app/navigation.js
@@ -15,20 +15,23 @@ document.addEventListener("DOMContentLoaded", () => {
     //When you mouseover/click, menu appears
     navList[i].addEventListener("mouseout", () => {
       stillHere = 1;
-      navAbsolute[i].classList.remove("navAbsoluteActive");
-      setTimeout(() => {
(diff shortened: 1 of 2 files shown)


fcaps:
not super beautiful fixed, but i have to rewrite it otherwise to be resilient to all kinds of issues.
Example:
- what if wordpress just sends some malformed data?
- what if the cache-file is malformed
- etc.

Brutus5000 (approved):

PR #460, opened 2023-11-12
Move clientRelease url generator

### LINKED ISSUE #418: Remove usage of github-api
Github API has a bug in it's _request functionality which makes it impossible to catch some kinds of errors in the returned promise of Repository.getRelease call. In our case we could not catch the error thrown when the github api returned a 403 due to rate limiting. This then can prevent start up of the application.

Due to this we should remove usage of github-api as the underlying issue has not been resolved. see https://github.com/github-tools/github/issues/439.

fcaps:
closes #418


public/js/app/play.js
```diff
@@ -0,0 +1,54 @@
+import {Octokit} from "https://esm.sh/@octokit/core";
```
Brutus5000, line 1: Looks like this was not in the package, but directly downloaded. Why?

fcaps: Maybe you missed that this file is executed in a browser. Just used the recommended way for browsers.

Sheikah45 (commented):
My main concern would just be that if someone exceeds the GitHub API rate limit then this would fail when querying for the latest but with the fallback that should be handled I believe and it is run client side so won't affect the server.

I would also add /latest to the fallback url so that the user is taken directly to the latest release.

fcaps:
good point with the latest, i'll change this.
Yeah, server is not affected.

public/js/app/play.js
Sheikah45: I guess the only potential discomfort I have here is about the presence of the script at the url and potential for injection.

Although to get around this we would have to use some bundler like vite or webpack correct

fcaps:
since we are only hitting the github-api on click, we should not be the origin of a rate-limit  (if not a developer and clicking the  button multiple times^^).
there is a potential issue if the github-api is slow (timeout reached), the user will get the fallback delayed, but at this point it's better than having a container not starting.

public/js/app/play.js
fcaps: even if we would have a builder, it will not change the potential harm.
if someone gets access to the repository (to any package we use), we can't do anything about it (forking and self-hosting would be an option), but since it's an official [sdk](https://github.com/octokit/octokit.js) from github, i don't have any discomfort.
there are other "security" issues looking into the code.

i would say, if we don't trust github or [esm](https://github.com/esm-dev/esm.sh), we should also not trust cloudflare or google-tagmanager/analytics^^

The github client used in node  was even more critical, since it could execute code in the container, this combined with node running as root and no checksum checks (already fixed with `--frozen-lockfile` option and the new Dockerfile) could have been more dramatic than just me ranting and creating a PR to fix that.

We also have to balance security, if the site would get compromised, they could only get a potential login-password, or impersonate FAF to trick the user to download/install/accept something that is not from us, if the browser-security is set to be wide open.

i would "trust" esm and github (maintainers of the repo), but maybe i am missing something here.

Sheikah45: I did not realize it was a github official sdk

fcaps: Was also surprised by the name

Sheikah45 (approved):

PR #461, opened 2023-11-12
Remove lobby_api querying

Brutus5000:
This is unused since new website (was bugged anyway) and currently stops moving the website to Kubernetes.


fcaps:
there was a [comment ](https://github.com/FAForever/website/commit/9f6d560ec7d0e1051bf66d964b41886a03134fbd)from @JaviTrek  
```
### lobby_api
reinstated after it seems it is actually being used and is not some useless piece of code. It provides users with a mirror of the users, their maps and lobbies in the FAF client.
```

But nonetheless, it's broken.
Can you also remove the `LOBBY_API_URL` in .env.example, env.faf-stack and in the faf-stack repo? 
Also the readme can be updated `The website has dependencies to`

fcaps:
lgtm :) thx for the update

fcaps:
you know where it was used? in the old game-client?

Brutus5000:
It used to show the number of current players online and games played. At some point there was also a world map where we showed where the players are.

Brutus5000:
But this could be brought back with a proper grafana chart integration maybe.

fcaps:
i see, was curios where this was used, did not find any ref in the game-client

PR #462, opened 2023-11-12
enable global error-handler again + some error-logging improvements

fcaps:
The global error-handler was disabled due some missing parameters (god knows i like this kind of magic).
Improved the logs to be more "production" friendly.
image

### DIFF
--- a/express.js
+++ b/express.js
@@ -322,7 +322,12 @@ setInterval(() => {
 app.use(function (req, res) {
   res.status(404).render('errors/404');
 });
-app.use(function (req, res) {
+app.use(function (err, req, res, next) {
+  console.error('[error] Incoming request to"', req.originalUrl, '"failed with error "', err.toString(), '"')
+  if (res.headersSent) {
+    return next(err);
+  }
+
   res.status(500).render('errors/500');
 });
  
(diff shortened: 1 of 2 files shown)


express.js
```diff
@@ -322,7 +322,12 @@ setInterval(() => {
 app.use(function (req, res) {
   res.status(404).render('errors/404');
 });
-app.use(function (req, res) {
+app.use(function (err, req, res, next) {
+  console.error('[error] Incoming request to"', req.originalUrl, '"failed with error "', err.toString(), '"')
+  if (res.headersSent) {
+    return next(err);
+  }
```
Brutus5000, line 329: What does this do?

fcaps: there is an issue if the error is thrown after the server already sent the http-header, in this case it's better to let express handle the error, see https://expressjs.com/en/guide/error-handling.html
There are some pending issues with the general approach.. how to handle errors if the caller expects json, js etc.

fcaps: if you missed it, this function was never called (thus no error handling on production) bc. of express expecting 4 parameters:
`Define error-handling middleware functions in the same way as other middleware functions, except error-handling functions have four arguments instead of three: (err, req, res, next)`

Brutus5000 (approved):

Issue #454, opened 2023-11-11
Replace setInterval


fcaps:
using `setInterval` can be a huge issue described in multiple articles.
just search "nodejs why setInterval is bad".
a common solution is something like https://www.npmjs.com/package/node-cron, or everything that is based on https://nodejs.org/api/events.html#class-eventemitter

fcaps:
this issue is more critical than thought, i forgot to shutdown my pc (and containers) for multiple hours, and the container was doing weird stuff.

steps to reproduce:
- start the container
- maybe kill the wordpress instance, so it will create issues
- let it run for some time (depends on your system)
- try to stop it -> not responsive
- kill it -> killed, but somehow not correctly
- start the same container again -> not starting
- wait some time
- start it again -> working
 
Since the website also has issues on prod where we cannot start it after a crash, i guess it's not a local issue. 

I would guess it's all the setInterval creating a memory/thread mess, but could be wrong.

fcaps:
update:
looking into this even more, i suspect having a "real" cache mechanism is even more useful than just updating a file every x mins.


old:
`server starts -> query endpoints of wp/java-api -> saves file on local disk`

new:
```
browser hits endpoint  ->
   -> cache is not stale -> deliver 
   -> cache is stale -> query endpoint -> safe cache -> deliver
```

Some upside with this approach:
- no need to have a cronjob running on the server (no setInterval/timeout)
- container can start without calling endpoints on java-api/wordpress
- no calls if there is no need to update (if there is no user)

downsides:
- one user is the "cache-looser" (long http request time)
- need a mutex/flock to not call the endpoint multiple times if cache is stale
- some may complain about memory usage (<10MB)
- some may complain about no file to look at 
- maybe more


packages that can be used:

https://github.com/node-cache/node-cache
maybe a mutex? but have to check it => https://stackoverflow.com/questions/64899813/does-node-cache-uses-locks

PR #463, opened 2023-11-12
Refactoring news-routes to be simpler and not using setInterval

fcaps:
This is related to #454

1. there was no reason to go the complicated way and render news in the browser
2. there was no reason to use the setInterval  in `getNews.js` `router.get(/news/${page})` -> replaced with `router.get(/news/:slug)`
3. updated the news-slug to be actual url safe, but to not break any public links there is a fallback + redirect 
4. started to use routers `app.use("/news", newsRouter)` to the `express.js `is not cluttered with code

Since the setInterval was adding so many routes to express, this has a good chance to be memory-leaking piece of code


Sheikah45:
Just a quick note on your first point that we have to make sure the newshub still works with the client. As the news is just the rendering of the news page.

From the code I suspect you mean the individual articles but wanted to make sure.

fcaps:
did not touch the http://localhost:8020/newshub, so changes will not affect it.
beware: the direct.faforever.xyz is not working correctly, so newshub will also not work if you test on faforever.xyz.

xyz will not have the `newshub_externalLinkUrl` in the items, see:

https://direct.faforever.com/wp-json/wp/v2/posts/?per_page=10&_embed&_fields=_links.author,_links.wp:featuredmedia,_embedded,title,newshub_externalLinkUrl,newshub_sortIndex,content.rendered,date,categories&categories=283

vs 
https://direct.faforever.xyz/wp-json/wp/v2/posts/?per_page=10&_embed&_fields=_links.author,_links.wp:featuredmedia,_embedded,title,newshub_externalLinkUrl,newshub_sortIndex,content.rendered,date,categories&categories=283

Brutus5000 (approved):
I don't really follow all changes, but we'll see on test

PR #469, opened 2023-11-17
removed /client route, it was not working since a long time

fcaps:
it was also not working before my change to the github-downlaod stuff. 
also..  don't get the use of a download link redirected via the server (maybe tracking?), but hey, who knows

### DIFF
--- a/express.js
+++ b/express.js
@@ -214,12 +214,6 @@ app.post('/account/activate', require(routes + 'account/post/activate'));
 app.get('/account/checkUsername', require('./routes/views/checkUsername'));
 app.get('/password_resetted', require(routes + 'account/get/requestPasswordReset'));
 app.get('/report_submitted', require(routes + 'account/get/report'));
-// Download Client
-app.get('/client', (req, res) => {
-  let locals = res.locals;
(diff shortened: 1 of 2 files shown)


Brutus5000 (changes requested):
The corresponding outdated a-element in index.pug should be removed too

fcaps:
> The corresponding outdated a-element in index.pug should be removed too

good catch, was not aware of it, maybe bc. it was already commented out xD

express.js
```diff
@@ -215,11 +215,6 @@ app.get('/account/checkUsername', require('./routes/views/checkUsername'));
 app.get('/password_resetted', require(routes + 'account/get/requestPasswordReset'));
 app.get('/report_submitted', require(routes + 'account/get/report'));
 // Download Client
```
Sheikah45: This comment can be removed as well correct?

fcaps: done

fcaps: we need to cleanup the express.js so bady, maybe if i got the energy for it.

Sheikah45 (approved):

PR #472, opened 2023-11-18
#451 move to session files

### LINKED ISSUE #451: Express session-store memory-leak
starting the docker image with "NODE_ENV=production" was logging something we should pay attention to.
```
Warning: connect.session() MemoryStore is not
designed for a production environment, as it will leak
memory, and will not scale past a single process.
```

could be fixed with something like [this](https://github.com/roccomuso/memorystore), but didn't had the time to dive into it.

fcaps:
closes #451 

Changes:
- moving from memory to file-stroage
- removed "normal" cookie -> it's a session cookie now (close the browser, losses the session)
- removed deprecated cookie-parser middleware, it's integrated in express-session
- sessions will not be lost if the container getting restarted (but on every deploy)

Notes:
- sessions are encrypted
- if the file-system is not "stable" , we may have to increase retries.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,3 +46,6 @@ public/styles/css/*
 .env
 
 public/js/*.js
+
+#Sessions
+sessions
--- a/express.js
+++ b/express.js
@@ -3,7 +3,7 @@ const express = require('express');
 const showdown = require('showdown');
 const passport = require('passport');
 const session = require('express-session');
-const cookieParser = require('cookie-parser');
+const FileStore = require('session-file-store')(session);
 const bodyParser = require('body-parser');
 const flash = require('connect-flash');
 const fs = require('fs');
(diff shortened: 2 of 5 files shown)


sessions/.gitkeep
Sheikah45, line 1: Why do we need the sessions folder in git? Does express not create it if it doesn't already exist?

Sheikah45 (commented):
Would the sessions just be stored in /sessions so we would have to mount a volume or would it be okay if they are wiped on container restart?

sessions/.gitkeep
fcaps: > Would the sessions just be stored in /sessions so we would have to mount a volume or would it be okay if they are wiped on container restart?

they are in-memory right now, so i guess it's okay if they got wiped. (PS: container restart will keep the sessions)

> Why do we need the sessions folder in git? Does express not create it if it doesn't already exist?

it's a permission thing, we can let node (the container) create it, but then permission are kinda fucked locally for development, it's a small price for not doing the whole setUID/GID for every operating system. but, i am used to it and can remove the folder.

Sheikah45: Yes I meant container recreation. And why does the local developer need to care about the permissions of the folder?

fcaps: he cannot delete the sessions.

fcaps: PS: we already have this issue with node_modules.

Sheikah45: Yes I was mostly curious if it required any changes to the server setup

fcaps: no, everything can stay as it is. the image don't need any mount-point.

fcaps: there is the possibility that i fucked something up (like a side-effect), but i don't now the app good enough to see what is missing/wrong. maybe i fucked the flash messages by removing the cookie-parser?
The code is looking so dirty, that i often don't know if i fucked something up or not.

What i tested:
i could login, used my token and logout

Brutus5000 (approved):
I'm ok with this for now, but that still limits us to 1 instance which is kind of an antipattern of what we try to achieve with the infrastructure side (kubernetes)

sessions/.gitkeep
Brutus5000: We'll have to do a larger testing session before the next release I guess

PR #481, opened 2023-11-20
added isAuthenticated middleware

### LINKED ISSUE #479: Redirect User After Login
if you go to a restricted site (account, leaderboard etc.) you will be redirected to login and after that to the "/" homepage.
ideally we should try to redirect the user to the original site after login, this is not possible and should also not be done in every scenario.

 Acceptance criteria:
- xhr/json request are not getting redirected at all (401 with error-message)
- post/put or any "updating" action/page will not be redirected after login
- somehow the ability to "configure" where to redirect after login per route. maybe middleware.isAuthenticated(redirectTo = '/account')

fcaps:
closes #479 
Redirecting after login is now possible, but i opted to not directly add this to every route.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "request": "2.88.2",
     "session-file-store": "^1.5.0",
     "showdown": "^2.1.0",
+    "supertest-session": "^5.0.1",
     "url-slug": "^4.0.1"
   },
   "devDependencies": {
--- a/routes/middleware.js
+++ b/routes/middleware.js
@@ -34,3 +34,21 @@ exports.username = function(req, res, next) {
 
   next();
 };
+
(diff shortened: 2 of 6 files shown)


Brutus5000:
Uh we got a merge conflict.

> but i opted to not directly add this to every route

If I understand the correctly, if you don't define anything you get back to the original url. Is that right? In that case, where else would we need to add this to routes?

fcaps:
> If I understand the correctly, if you don't define anything you get back to the original url. Is that right? In that case, where else would we need to add this to routes?

yeah, originUrl is the "current" path.
already added the middleware in PR https://github.com/FAForever/website/pull/485 to the other routes.

Brutus5000 (approved):

Issue #482, opened 2023-11-20
Leaderboard does not show progress or failures 


fcaps:
Since we have a cache-looser, it can take a while to load the entries.
This will result in a page, that just shows a empty list for a couple of seconds without telling the user to wait.

Acceptance Criteria:
- opening the leaderboard will show a loader if a request in due
- changing the leaderboard to "2v2" etc. will also show a loader
- failed request will be handled and show the user something like a error message

non-goals:
- fixing the leaderboardRepository if the token is invalid (currently creating a 503 on stale token)

PR #483, opened 2023-11-20
increased ttl to relax #482

fcaps:
related #482 

4 hours cache is better than having an unresponsive leader board.
change my mind

### DIFF
--- a/lib/LeaderboardService.js
+++ b/lib/LeaderboardService.js
@@ -1,4 +1,4 @@
-const leaderboardTTL = 60 * 60 * 4 // 4 hours ttl as a relaxation for https://github.com/FAForever/website/issues/482
+const leaderboardTTL = 60 * 60  // 1 hours ttl as a relaxation for https://github.com/FAForever/website/issues/482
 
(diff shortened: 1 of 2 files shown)


Brutus5000:
When I tested it on the freshly deployed website the loading time wasn't even 10 seconds. 4hours TTL is brutal. How about 1 hour?

fcaps:
> When I tested it on the freshly deployed website the loading time wasn't even 10 seconds. 4hours TTL is brutal. How about 1 hour?

lowered to 1h, this will be gone with m2m anyways.

fcaps:
btw. you got any traffic-numbers for me? is this site called 10 or 1000 times per day?

Brutus5000:
Uh such a long time ago that we had measured with plausible. But it was surprisingly high. I think more than 1000 per day.

fcaps:
GoogleAnalytics? can see the call to GA

Brutus5000:
Due to our privacy policy plausible was the maximum we want to use (or anything selfhosted).

Issue #480, opened 2023-11-20
Bootstrapping The Application


fcaps:
Separate the bootstrapping and the actual "listen" call of the application would benefit development and testing.

Actual behavior:
node is starting the express.js with cronjobs, routes, middlewares and the listen to the port.

Desired:
Routing (controllers), middlewares, configs and cronjobs can be bootstrapped without starting the webserver.

benefits:
- cleaner express.js (it is getting smaller, but it still does way to much)
- testing will be improved due the possibility to start the app without starting crons or the webserver
- reduced merge conflicts due the separation of concerns (and smaller files)
- developing new features will now "remember" the dev to think about where to put code.

Issue #233, opened 2020-05-24
Report form autocomplete = broken


Viba:
Users are unable to report other users due to the autocomplete not autocompleting properly.

e.g. trying to report "Kalvirox" it returns results for the first two and three letters like below (with weird spaces), then nothing. If the form doesn't find a valid player it doesn't allow filing a report. This is quite important to moderation.

![image](https://user-images.githubusercontent.com/3510353/82757934-1b0e8300-9dec-11ea-8daa-ec0caf2e7de1.png)
![image](https://user-images.githubusercontent.com/3510353/82757940-206bcd80-9dec-11ea-9b92-c70ec17be0da.png)

Brutus5000:
It works but it is broken by design.
![image](https://user-images.githubusercontent.com/6843462/82761429-c58da280-9dfa-11ea-9317-3b44a54b12bc.png)

I looked up the code and it seems to work like this:
It only collects player names of the players who played a game in the last 24 hours at the moment of startup and never refreshes it.

1-alex98:
I will just check if I am reportabe and if not I can do what I want XD @Brutus5000 when is the next restart?

Brutus5000:
Looks like it crashed 11 minutes ago and was restarted

1-alex98:
Well it updates the names every 15 minutes

norraxx:
We shouldn't search inside the usernames with `LIKE %{}%`. It's very expensive operation.
`LIKE {}%` is much much faster.

Rackover:
You can submit a report even if the autocomplete does not include it now, so is that still a problem?
The suggestions when writing someone's name are incomplete because they only include players from the last 24 hours for performance/caching reasons, but this could be changed and/or extended if necessary - ultimately as long as we rely on caching it will result in passing a huge list of names  to the client/browser and making the page more heavy (or we could kill caching and fetch the names from the API as they are typed, realtime, but not sure how to do that with awesomplete)

Brutus5000:
Are we sure it is refreshed once per day? I think I didn't find anything related back then.

Rackover:
Every 15 minutes normally! https://github.com/FAForever/website/blob/develop/express.js#L293
Script is there : https://github.com/FAForever/website/blob/develop/scripts/getRecentUsers.js

Now only way to see if it actually runs on prod is to check website logs on prod. Locally it appears to work, but test server has no recent games so no players get updated on it! Also 15 minutes might be a bit long...

[...]

PR #487, opened 2023-11-21
wordpress service + event-emitters

### LINKED ISSUE #454: Replace setInterval

fcaps:
closes #454
related to #480


lib/WordpressRepository.js
```diff
@@ -0,0 +1,103 @@
+const {convert} = require("url-slug");
+
+class WordpressRepository {
+    constructor(wordpressClient) {
+        this.wordpressClient = wordpressClient
+    }
+    async fetchNews() {
+        let response = await this.wordpressClient.get('/wp-json/wp/v2/posts/?per_page=100&_embed&_fields=_links.author,_links.wp:featuredmedia,_embedded,title,content.rendered,date,categories&categories=587')
```
Brutus5000, line 8: I think the categories url was configurable via env?
But I see so many categories now, I'm no longer sure.
Maybe we should at least put the ids as const variables at the top of the file for better naming in the actual urls?

fcaps: was just copy&pasta, https://github.com/FAForever/website/pull/487/files#diff-7e4783341040a4ba6619cd0157ccb8d361c450d4ac84148a2fe589414db924ddL22

maybe we can optimize it later? there are for sure other problems, like "options/fields" not used...

routes/views/account/get/report.js
```diff
@@ -101,23 +101,7 @@ exports = module.exports = function (req, res) {
       }
 
       locals.reportable_members = {};
-      const recentMembersPath = 'members/recent.json';
-      if (fs.existsSync(recentMembersPath)){
```
fcaps, line 105: we got this covered in #233, when the m2m is ready

scripts/cron-jobs.js
```diff
@@ -1,16 +1,27 @@
...
+    wordpressService.getNewshub(true).then(() => successHandler('getNewshub')).catch((e) => errorHandler(e, 'getNewshub'))
+    wordpressService.getContentCreators(true).then(() => successHandler('getContentCreators')).catch((e) => errorHandler(e, 'getContentCreators'))
+    wordpressService.getTournamentNews(true).then(() => successHandler('getTournamentNews')).catch((e) => errorHandler(e, 'getTournamentNews'))
+    wordpressService.getFafTeams(true).then(() => successHandler('getFafTeams')).catch((e) => errorHandler(e, 'getFafTeams'))
+}
+
+module.exports = async () => {
+    await warmupWordpressCache()
+    
+    const wordpressScheduler = new Scheduler('createWordpressCaches', warmupWordpressCache, 60 * 59 * 1000)
```
fcaps, line 25: i'll make all the "magic" numbers into a config once it is somewhat stable and does not change so often.
don't wanna add env-vars directly

templates/views/newshub.pug
```diff
@@ -19,8 +19,6 @@ html(lang='en')
 
     //- Customise the stylesheet for your site by editing /public/styles/site.sass
     link(href="/styles/css/site.min.css?version="+Date.now(), rel="stylesheet")
-    script(async, defer, data-domain="faforever.com", src="https://plausible.faforever.com/js/plausible.js")
-    script(src="https://kit.fontawesome.com/33e7e258d3.js" crossorigin='anonymous')
```
fcaps, line 23: plausible http 404
kit: http 403

don't know why this was loaded, but the newshub still looks okay

Brutus5000: Plausible is currently disabled on the server.

Brutus5000 (approved):

PR #497, opened 2023-11-26
move to webpack for cache busting and other benefits

### LINKED ISSUE #496: Enable cache busting for assets on deploy
What happened:
a route got removed from production "js/app/members/newshub.json"
this route was called from a js file that got cached in the reverse-proxy.

to overcome this issue, assets should be loaded with a version qyery-parameter like "https://faforever.com/js/app/newshub.js?v=2"

fcaps:
closes #496 

this is the first step to move to webpack.
- moved all "public/js" files into a new directoy "src/frontend", so we can run webpack and grunt without issues
- added a new static-asset entry for "dist" in express
- removed all related grunt tasks for js files.
- removed empty "account.js"
- moved octokit to node_modules

known issues:
- development is running a "prod" configuration for webpack, not optimal, but it works for now
- webpackManifestJS could be lazy-loaded, or hot-reloaded when on development


Sheikah45 (approved):
Although I am curious why webpack and not vite.

Purely from an academic standpoint.

fcaps:
webpack is well known, but thb. i was not evaluating builders.
i know everything i need right now is covered by webpack, and for other developers it would be easy to understand whats happening. 
What i heard from vite is they are focused more on dev, and not so on prod, but can't really give you a example.

fcaps:
i also guess moving from webpack to vite is easier than from grunt to vite. 
cleaning up the code was harder, than actual switch. never knew what is going on, and had to go in deeply to see if a file was used/working.

Sheikah45:
Understood, I was mostly curious as a framework I have been using made the switch from webpack to vite so was just curious if it was a general movement or just a framework preference.

Issue #484, opened 2023-11-20
Account Registration checkUsername


fcaps:
when you create an account, the check if a username is already taken fails

1. some regex is not working (error in the browser)
2. the api needs a m2m token for checkUsername.js

Brutus5000:
Registration was removed from the website

PR #513, opened 2023-12-02
Develop 484

beckpaul:
Addresses point 1 of #484 
Escaping `-` clears the error from browser
Includes some trimming of trailing white-space


fcaps:
looks like `data-remote, data-minlength-error` etc. are not used anymore, maybe they got refactored at some point.

fcaps (approved):

Issue #515, opened 2023-12-02
/account/password/confirmReset has no validation on token/username


fcaps:
this is how the url is actually called: `/account/password/confirmReset?username=minion&token=xxx`, but you can call it without the parameters and the form will still show up. 

at least i would have the input validated/escaped and some warning showing that the link is not correct. (and not showing the form)

fcaps:
@beckpaul i'll look for an option to let you "assign" issues, have to ask the admin for some kind of contributor role.

PR #516, opened 2023-12-03
Parameter validation for /account/password/confirmReset #515

beckpaul:
address #515 via redirect to request password reset and flash a message.

Bit more whitespace removal just while grepping around.

Change to .gitignore - mostly for my sanity as i use asdf to control dev envs.

Might have some overlap with #514

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -12,7 +12,9 @@ services:
       context: .
     networks:
       -  faf-stack
-    
+    logging:
+      driver: "json-file"
+
 networks:
   faf-stack:
     name: faf-stack_faf
--- a/src/backend/routes/views/account/get/confirmPasswordReset.js
+++ b/src/backend/routes/views/account/get/confirmPasswordReset.js
@@ -11,14 +11,19 @@ exports = module.exports = function (req, res) {
     locals.username = req.query.username
     locals.token = req.query.token
 
-    const valid_req = (!locals.token || !locals.username)
+    // Ensure token and username are present
+    const validateParamPresence = (token, username) => {
+      if (!token || !username) {
+        return "Invalid request"
+      } else {
+        return null
+      }
+    }
 
-    if (valid_req) {
-      let flash = {}
-      flash.class = 'alert-danger'
-      flash.messages = [{ msg: 'Invalid reset link' }]
-      flash.type = 'Error!'
-      res.redirect('/account/requestPasswordReset')
+    const errorMsg = validateParamPresence(req.query.username, req.query.token)
+
+    if (errorMsg) {
+      res.redirect('/account/requestPasswordReset?flash=' + errorMsg)
     } else {
       res.render('account/confirmPasswordReset')
     }
--- a/src/backend/routes/views/account/get/confirmPasswordReset.js
+++ b/src/backend/routes/views/account/get/confirmPasswordReset.js
@@ -1,33 +1,49 @@
+const axios = require('axios')
+const appConfig = require('../../../../config/app')
 const { validationResult } = require('express-validator')
 exports = module.exports = function (req, res) {
-    const locals = res.locals
-
     const errors = validationResult(req)
 
     if (!errors.isEmpty()) {
-        const flash = {}
-        flash.class = 'alert-danger'
-        flash.type = 'Error!'
-        flash.messages = []
-        errors.array().forEach((e) => {
-            flash.messages.push({ msg: e.msg })
-        })
-
-        const buff = Buffer.from(JSON.stringify(flash))
-        const data = buff.toString('base64')
-
-        return res.redirect('/account/requestPasswordReset?flash=' + data)
+        return renderRequestPasswordReset(req, res, errors)
     }
 
-    // locals.section is used to set the currently selected
-    // item in the header navigation.
-    locals.section = 'account'
+    res.render('account/confirmPasswordReset', {
+        section: 'account',
+        formData: req.body || {},
+        username: req.query.username,
+        token: req.query.token
+    })
+}
 
-    locals.formData = req.body || {}
+const renderRequestPasswordReset = async (req, res, errors) => {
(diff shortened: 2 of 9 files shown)


src/backend/routes/views/account/get/confirmPasswordReset.js
```diff
@@ -1,16 +1,40 @@
...
+        const flash = {}
+        flash.class = 'alert-danger'
+        flash.messages = [{ msg: errorMsg }]
+        flash.type = 'Error!'
+
+        const buff = Buffer.from(JSON.stringify(flash))
+        const data = buff.toString('base64')
+
+        return overallRes.redirect('/account/requestPasswordReset?flash=' + data)
+    } else {
```
fcaps: else not needed when returning in a error-case

src/backend/routes/views/account/get/confirmPasswordReset.js
```diff
@@ -1,16 +1,40 @@
 exports = module.exports = function (req, res) {
     const locals = res.locals
 
+    const overallRes = res
```
fcaps: is this a express best practice? or did i miss something?

src/backend/routes/views/account/get/confirmPasswordReset.js
```diff
@@ -1,16 +1,40 @@
...
 
-    const flash = null
-
     // Render the view
     locals.username = req.query.username
     locals.token = req.query.token
-    res.render('account/confirmPasswordReset', { flash })
+
+    // Ensure token and username are present
+    const validateParamPresence = (token, username) => {
```
fcaps: i know this works, but should we force to use "express-validator" for validations?
what do you think? 
```JavaScript
const { query, validationResult } = require('express-validator')
#....

 query('username')
        .isString().withMessage('Username missing')
        .trim()
        .notEmpty()
```

tests/integration/accountRouter.test.js
```diff
@@ -22,6 +21,17 @@ describe('Account Routes', function () {
...
 
+    test('responds with OK to provided parameters', async () => {
+        const response = await testSession.get('/account/password/confirmReset?username=turbo2&token=XXXXX')
+        expect(response.statusCode).toBe(200)
+    })
+
+    test('redirect to reset request page if missing parameters with flash parameter', async () => {
+        const response = await testSession.get('/account/password/confirmReset')
+        expect(response.statusCode).toBe(302)
+        expect(response.headers.location).toContain('/account/requestPasswordReset?flash=')
```
fcaps: if you follow the request, you could expect for the "invalid request" and get the other code covered.

fcaps (changes requested):
see the questions above.

src/backend/routes/views/account/get/confirmPasswordReset.js
beckpaul: Unsure about best practice, but it is how flashes on redirect are done in many other places in the code base.

src/backend/routes/views/account/get/confirmPasswordReset.js
beckpaul: Definitely cleaner and better option - will make the swap over.

src/backend/routes/views/account/get/confirmPasswordReset.js
fcaps: i think this was bad practice when they had a api-request and called the result "res", so the variable was overwritten.
something like this gem:
```javascript
 const overallRes = res

    request.post({
        url: process.env.API_URL + '/users/resyncAccount',
        headers: { Authorization: 'Bearer ' + req.requestContainer.get('UserService').getUser()?.oAuthPassport.token }
    }, function (err, res, body) {
        if (err || res.statusCode !== 200) {
            error.parseApiErrors(body, flash)
        } else {
        // Successfully account resync
            flash.class = 'alert-success'
            flash.messages = [{ msg: 'Your account was resynced successfully.' }]
            flash.type = 'Success!'
        }

        overallRes.render('account/confirmResyncAccount', { flash })
    }
    )
```

they could have just called it "resyncResponse", but they decided to override the "res" because of reasons..

fcaps: we lost a lot of history due my refactoring, but most of the code is old and was just a copy

beckpaul: Ran into a lot of issues with that syntax for passing the flash to the redirect:

`overallRes.render('account/confirmResyncAccount', { flash })`

Will see if there's a better way to do this and potentially adapt it in other places.

beckpaul: You are correct here - overallres was bad practice, just a remnant while copying over using params for flash.

tests/integration/accountRouter.test.js
beckpaul: Does not appear that following the request is configured as a standard option nor done as a current standard.

I would think i could test it using `response.res.locals` but that's undefined

Reading from the text response you could check for `<ul class=\"flash-errors\">` but thats present regardless of the case and is empty.

Potentially readable from session cookies but hit a wall there. 

Will revisit this tomorrow.

fcaps: the website is broken everywhere, created a solution for your problem:
https://github.com/beckpaul/faf-website/pull/1/files

in general: flash messages are broken, they often have not the right structure (someone changed the layout)
tests: i restricted the tests and disallowed non localhost requests, so every request done needs to have a mock if not to localhost

docker-compose.yml
```diff
@@ -12,7 +12,9 @@ services:
       context: .
     networks:
       -  faf-stack
-    
+    logging:
+      driver: "json-file"
```
fcaps, line 16: is this still needed?

fcaps (approved):

docker-compose.yml
beckpaul: Will check and remove in next pr if not.

fcaps: already removed it, so you need to add it again... sry :)

PR #519, opened 2023-12-03
reactivate clan-list/view with the new m2m client

### LINKED ISSUE #459: clan/{id} was removed, but game-client still uses it
The game-client (_chat -> right click on a user -> visit clan page_) is opening the clan-page with `https://clans.faforever.com/clan/{id}`, but this route was removed.
There was some redirect in the code, but it got also removed with #410 .
```
// Compatibility
app.get('/clan/*', function (req, res){
  const id = req.path.split('/').slice(-1)[0];
  res.redirect('/clans/see?id='+id);
});
```

there is an [issues ](https://github.com/FAForever/downlords-faf-client/issues/3035)open, but i would not opt to replace `id` with `tag`, just update the route `clans` with an `s`.

Till implemented in the game-client, we can enable the redirect again (when clans are working).

fcaps:
closes #459 
related #445 

as a part of the refactoring i had to remove the old code (it was not maintainable anymore) and updated it to work with the new m2m client, services and repository pattern.

The single view: (the leave clan is not implemented and will be done in another ticket)
image

List view:
image

The filter is instant (no backend call) and is bound to filter tag, name or leader.

Caching for clans/clan is 60min and should be purged if we add the clan-management again, so we purge only if needed.
There is also the possibility to "warmup" the clan cache, but this could be handled in another ticket.


fcaps:
@beckpaul you wanna try reviewing this?

fcaps:
i know i should cover the error-cases a bit better, but somehow my joy writing tests has limits...

Issue #520, opened 2023-12-03
Clan cache warmup


fcaps:
as described in PR https://github.com/FAForever/website/pull/519
clans (getAllClans()) could benefit a warmup like we did it for `leaderboardCacheCrawler.js`.
warmup a single clan is too much, since the call is ~500ms and this i can take a cache-looser.

Acceptance Criteria:
- cache will be generated at server start/restart
- will be scheduled to refresh it every (ttl - 1min), currently this would be 59min
- should not run in parallel with other JavaApi calls if possible (ddos ourself if the list of crawlers getting bigger)
- should not crash the server if the JavaApi is not reachable or and exception is happening.
- should log if a cache was getting refreshed or when something failed

PR #526, opened 2023-12-06
Clan cache warmup #520

beckpaul:
Address #520 for the most part.

## Criteria:
*cache will be generated at server start/restart*

- Runs on start along with other jobs

*will be scheduled to refresh it every (ttl - 1min), currently this would be 59min*

- TTL configured is 60 min in clanService - job runs every 59

*should not run in parallel with other JavaApi calls if possible (ddos ourself if the list of crawlers getting bigger)*

- I believe mutex takes care of this here:
```
from src/backend/services/clanService.js

if (this.mutexService.locked) {
        await this.mutexService.acquire(() => {
        }, this.lockTimeout)
    return this.getAll()
}
```
*should not crash the server if the JavaApi is not reachable or and exception is happening.*

- Exceptions are handled the same way as leaderboards with logging and catching

*should log if a cache was getting refreshed or when something failed*

- Error handler and success handler print when caches updates/fails.

### DIFF
--- a/src/backend/AppKernel.js
+++ b/src/backend/AppKernel.js
@@ -20,6 +20,7 @@ const leaderboardRouter = require('./routes/views/leaderboardRouter')
 const clanRouter = require('./routes/views/clanRouter')
 const accountRouter = require('./routes/views/accountRouter')
 const dataRouter = require('./routes/views/dataRouter')
+const { CacheService } = require('./services/CacheService')
 
 class AppKernel {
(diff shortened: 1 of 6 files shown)


Brutus5000:
Please use proper PR names as they appear in the changelog

fcaps:
you could use the names of the ticket aka "524-clan-transfer-ownership", there is even a create branch function in the ticket.

fcaps (approved):

PR #530, opened 2023-12-12
Internal: Add prettier and simplify eslint config

beckpaul:
Due to the size of the PR ill be addressing the relevant file changes/new files in comments below.

Main motivation behind this is to address the egregious number of white space issues and large trailing single lines.


.eslintrc
```diff
@@ -1,7 +1,34 @@
 {
```
beckpaul, line 1: Add prettier as a plugin for eslint - it gets run now when `yarn lint` is ran and `--fix` will also fix prettier issues

Not sure why there were 4 different `.eslintrc` files in different places but i've removed them and compressed the rules into one.

.prettierrc.yaml
```diff
@@ -0,0 +1,10 @@
+# See https://prettier.io/docs/en/options. most are set to default
```
beckpaul, line 1: Config for prettier - per comment: most are set to default. We can change this to JSON TOML or plain js if anyone has an aversion to yaml

package.json
```diff
@@ -34,9 +34,11 @@
     "css-loader": "^6.8.1",
     "dart-sass": "^1.25.0",
     "eslint": "^8.54.0",
+    "eslint-config-prettier": "^9.1.0",
```
beckpaul, line 38: Add prettier, the plugin to socket it in, and the config that prevents conflicts with eslint.

Essentially we keep eslint as a way to find and prevent bugs, while prettier is soley focused on code styling per:
https://prettier.io/docs/en/comparison

.eslintrc
fcaps: it's the same with .gitignore, there are multiple ways to configure things, globally or per directory.
i think we can work with overrides for now

fcaps:
Thx for the work!

not a fan of prettier per se, but i can work with it. 
most white-space issues are in other files than `.js`, that said, how do we fix them now?

beckpaul:
> Thx for the work!
> 
> not a fan of prettier per se, but i can work with it. most white-space issues are in other files than `.js`, that said, how do we fix them now?

Something like this: [eslint-plugin-pug](https://github.com/valpackett/eslint-plugin-pug)
will lint the inline scripts for the `.pug` files, not the "view" itself.

Benefit of prettier here is that we can add: [@prettier-plugin-pug](https://github.com/prettier/plugin-pug)
and that will lint the files as a whole and include styling. We can also add some pug specific rules if deemed necessary. List of possibilities: [here](https://prettier.github.io/plugin-pug/guide/pug-specific-options.html)

Went ahead and added just the prettier plugin in the next few commits and ran it with fix.
Take a look at the changes made to the scripts in `package.json` to make sure those are fine or if we want to break them out. 

Ideally, `yarn eslint ...` would entail also running prettier on our pug files but haven't gotten a proper config to make that happen so running it explicitly for the time being - will revisit this later today and see if i can clean it up.

package.json
```diff
@@ -51,6 +54,7 @@
     "load-grunt-tasks": "5.1.0",
     "nock": "^13.3.8",
     "octokit": "^3.1.2",
+    "prettier": "^3.1.1",
     "style-loader": "^3.3.3",
     "supertest": "^6.3.3",
     "typescript": "^5.3.2",
```
beckpaul, line 60: `typescript` Is this necessary?

fcaps: you can run `docker compose run website yarn npm-check` to check it, but i think this was needed for some webpack commands. not sure though

fcaps: just checked it, it is used by the `node-dependency-injection`, they used it to autowire services and i never had a chance to investigate on how to disable it.

fcaps:
```
website-website-1  |     Error: Cannot find module 'typescript'
website-website-1  |     Require stack:
website-website-1  |     - /code/node_modules/@typescript-eslint/typescript-estree/dist/convert.js
website-website-1  |     - /code/node_modules/@typescript-eslint/typescript-estree/dist/ast-converter.js
website-website-1  |     - /code/node_modules/@typescript-eslint/typescript-estree/dist/parser.js
website-website-1  |     - /code/node_modules/@typescript-eslint/typescript-estree/dist/index.js
website-website-1  |     - /code/node_modules/node-dependency-injection/dist/lib/Autowire.js
website-website-1  |     - /code/node_modules/node-dependency-injection/dist/lib/Loader/FileLoader.js
website-website-1  |     - /code/node_modules/node-dependency-injection/dist/lib/Loader/JsFileLoader.js
website-website-1  |     - /code/node_modules/node-dependency-injection/dist/lib/index.js
website-website-1  |     - /code/src/backend/dependency-injection/AppContainer.js
website-website-1  |     - /code/src/backend/AppKernel.js
website-website-1  |     - /code/src/backend/index.js

```

fcaps:
@beckpaul you think this can be merged?

beckpaul:
> @beckpaul you think this can be merged?

@fcaps Should be good to go  - only concern is if there's any other file types we would want to lint but that could come later since the only type that comes to mind is our CSS which is due for a refactor per another issue.

PR #532, opened 2023-12-15
Clan Ownership refactor

### LINKED ISSUE #524: [Clan] transfer ownership
this is the last function that needs to be refactored and enabled again.
take a look how the update/invite etc. are done, and apply it to transfer.

beckpaul:
Closes #524 

Adds a "make leader" button to the view clan page for clan leaders. Allows leaders to transfer leadership in clans utilizing the new standards and helper functions.


src/backend/services/ClanManagementService.js
```diff
@@ -36,6 +36,24 @@ class ClanManagementService {
         }
     }
 
+    async transferOwnership(newOwnerMemberId, clanId) {
+        const newOwnerUser =
+            await this.clanService.getClanMembership(newOwnerMemberId)
```
beckpaul: I'm unsure that this returns what I want it to:

need to patch in the new owner like:
```
leader: {
  data: {
      id: newOwnerId,
      type: 'player',
```
So i need the player ID retrieved from the membershipID. Not seeing any documentation for this but its based on how this was previously implemented + a bit of faith.

@fcaps if you can shed some light on this then it should be good for testing before merge

src/backend/routes/views/clans/transfer.js
```diff
@@ -0,0 +1,32 @@
+const { JavaApiError } = require('../../../services/ApiErrors')
+
```
beckpaul, line 2: Might want to add a validation back to this. We can assume that the single necessary param is here based on the view so likely not be a real concern

src/backend/services/ClanManagementService.js
fcaps: i had to look into it and thought i could directly fix it.
in short: the clan.members already got a playId, so we don't need the memebrship.
https://github.com/beckpaul/faf-website/pull/2

beckpaul:
Uploading Screen Recording 2023-12-30 at 13.59.33.mov…

beckpaul:
https://github.com/FAForever/website/assets/50498548/d8253927-e6e4-4257-9267-39c1e26d8a2e

fcaps (approved):

PR #534, opened 2023-12-30
add snowflakes

### LINKED ISSUE #533: Fix the wording of the rule regarding multiple accounts
The rule on multiple accounts on the website should be reworded:
>Do not abuse by using multiple accounts."

This currently implies that it's fine to have multiple accounts as long as you don't cause abuse.
Suggested to reword it to:  "Do not create more than one account", as that is the intended meaning.

fcaps:
closes #533

### DIFF
--- a/src/backend/templates/layouts/default.pug
+++ b/src/backend/templates/layouts/default.pug
@@ -293,6 +293,14 @@ html(lang='en')
                             a(href='https://forum.faforever.com/') FORUMS
 
             script(src=webpackAssetJS('navigation'))
+            script(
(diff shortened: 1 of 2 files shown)


fcaps:
snow

Askaholic:
Should it turn on/off automatically based on the date? That would probably be better than requiring a developer to merge a code change and have it get deployed each time.

fcaps:
> Should it turn on/off automatically based on the date? That would probably be better than requiring a developer to merge a code change and have it get deployed each time.

also thought about it, but then decided to keep it simple.
you can update this PR if you think this would be good to implement.

Brutus5000 (approved):

PR #535, opened 2024-01-07
Addition - Secondary account disclaimer

beckpaul:
Adds disclaimer to register page regarding multiple account operation and ownership as well as advise on recovery methods.

Screenshot 2024-01-11 at 14 57 30

### DIFF
--- a/src/backend/templates/views/account/register.pug
+++ b/src/backend/templates/views/account/register.pug
@@ -20,7 +20,7 @@ block content
                             href='/account/requestPasswordReset',
                             target='_blank'
                         ) Lost Account or Returning Player
-                p Creating and operating multiple accounts is prohibited.
+                    p Creating and operating multiple accounts is prohibited.
                 form(
                     method='post',
                     action='/account/register',
--- a/src/backend/templates/views/account/register.pug
+++ b/src/backend/templates/views/account/register.pug
@@ -16,11 +16,15 @@ block content
                 hr
                 h3 Lost account / returning player
                 p If you have a FAF account you can no longer access, please attempt to &nbsp
-                  a(href="https://help.steampowered.com/en/wizard/HelpWithLoginInfo?reset=1&issueid=408" target="_blank") reset your Steam password.
-                  | &nbspFor further assistance, contact the moderation team via&nbsp
-                  a(href=discordUrl target="_blank") Discord
-                  |  or the&nbsp
-                  a(href=forumUrl target="_blank") Forums.
+                    a(
+                        href='https://help.steampowered.com/en/wizard/HelpWithLoginInfo?reset=1&issueid=408',
+                        target='_blank'
+                    ) reset your Steam password.
+                    | &nbspFor further assistance, contact the moderation team via&nbsp
+                    a(href=discordUrl, target='_blank') Discord
+                    |
+                    | or the&nbsp
+                    a(href=forumUrl, target='_blank') Forums.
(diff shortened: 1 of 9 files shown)


src/backend/templates/views/account/register.pug
```diff
@@ -13,6 +13,10 @@ block content
                 br
                 +flash-messages(flash)
 
+                hr
+                h3 Lost account / returning player
+                p If you previously had a FAF account you no longer can get access to, please contact the moderation team via discord or forums, or see if you can reset your password using the Steam API. Note that creating and operating multiple accounts is prohibited, and creating a secondary account could lead to moderation action.
```
pumbas600: Slight rewording. What do you think?

```suggestion
                p If you previously had a FAF account you can no longer access, please contact the moderation team via Discord or the forums, or see if you can reset your password using the Steam API. Note that creating and operating multiple accounts is prohibited, and creating a secondary account could lead to moderation action.
```

beckpaul: Text with formatting and links removed:

If you have a FAF account you can no longer access, please contact the moderation team via Discord or the Forums, or attempt to reset your password on Steam.
Creating and operating multiple accounts is prohibited, and creating a secondary account will lead to moderation action.

Screenshot 2024-01-08 at 07 47 33

beckpaul: @pumbas600 
Added your suggestions and a few other tweaks to make it a bit less verbose.
Not 100% about including links for discord/steam since if those change then it will require someone to go in and edit them

Sheikah45: I would probably put the reset your password using steam first since that normally always works and is something the user can do on their own.

beckpaul: Screenshot 2024-01-08 at 10 48 04
 
 ```If you have a FAF account you can no longer access, please attempt to reset your Steam password. For further assistance, contact the moderation team via Discord or the Forums.```

pumbas600: > including links for discord/steam since if those change then it will require someone to go in and edit them

I've not properly looked through the pug files, but I imagine we're also using these links in the top right of the website? I believe you can create variables for these links and then use the variables in the pug files, rather than hardcoding it directly. That would mean if they needed to be changed, it could be done in one place :)

beckpaul: Good catch, probably should have been done a while ago. Did so in [2ac1962](https://github.com/FAForever/website/pull/535/commits/2ac1962bab4df46ccc4b13bb1c77371cd01bc0e7)

Might apply this to more links later in the day after seeing what else could benefit from this

fcaps: ping me if you need some input/merge, nice to see some work being done :)

BlackYps: "could lead to moderation action"? That sounds like we will only maybe enforce this. I would suggest either rewording it to "will lead to moderator action" or shortening the whole sentence to "Note that creating and operating multiple accounts is prohibited." The second half is a bit redundant

beckpaul: > "could lead to moderation action"? That sounds like we will only maybe enforce this. I would suggest either rewording it to "will lead to moderator action" or shortening the whole sentence to "Note that creating and operating multiple accounts is prohibited." The second half is a bit redundant

Didn't include the second line in my last 
`Creating and operating multiple accounts is prohibited, and creating a secondary account will lead to moderation action.`

I see the redundancy however and agree shortening is fine here.

fcaps (approved):

PR #536, opened 2024-01-08
Increase pill roundness and horizontal padding

pumbas600:
Hi :wave:

I noticed that the pill buttons on the website in a lot of cases weren't *quite* rounded enough to be a full pill button.

I also increased the horizontal padding on the buttons which I think improves their shape slightly.

| Before                                        | After                                                           |
|-----------------------------------|-----------------------------------------------|
|  Before Card |  After Card |
|  Before Landing |  After Landing |

### DIFF
--- a/public/styles/abstracts/_mixins.sass
+++ b/public/styles/abstracts/_mixins.sass
@@ -33,7 +33,7 @@
   border: 0px solid variables.$background-secondary
   background-color: variables.$button-inactive
   font-family: variables.$family-subtitle
-  border-radius: 20px
+  border-radius: 9999px
   font-size: 1.5em
   justify-self: center
   transition: 0.4s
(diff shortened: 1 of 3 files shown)


fcaps:
Hi @pumbas600,
and thx for your PR!

There is css "trick" for doing it without the 9999px  `border-radius: 25%/50%` looks quite similar, you think there are drawbacks?
Other changes looking good to be merged.

public/styles/abstracts/_mixins.sass
```diff
@@ -33,7 +33,7 @@
   border: 0px solid variables.$background-secondary
   background-color: variables.$button-inactive
   font-family: variables.$family-subtitle
-  border-radius: 20px
+  border-radius: 9999px
```
fcaps, line 36: border-radius: 25%/50%; looks similar

pumbas600: Percentages make it elliptical. While this is fine for some buttons, it becomes more noticeable for buttons that are longer, or not as tall.

| `25%/50%` | `9999px` |
|-------------|----------|
|  image | image |


While the `9999px` looks weird, it's just a massive number so that you never have to worry about the button being too large. To be honest, `99px` probably would've been sufficient too 😅

fcaps (approved):

PR #558, opened 2024-10-25
Rules text and styling preview

FredrikBergelin:
Updated rules text and styling for all markdown pages. More changes will follow, team want to review the current changes on staging environment.

Also snowflakes are set to start in December, not October.

// FreadyFish


Sheikah45:
The only thing I am curious about is why the paths need to be hidden instead of just using the test server. @Brutus5000

FredrikBergelin:
No reason, just me not doing my homework before I start macgyvering. Will fix

src/backend/templates/views/markdown/privacy.md
FredrikBergelin, line 1: Was not written with a Russian R (the one that looks like "P")

Brutus5000:
> The only thing I am curious about is why the paths need to be hidden instead of just using the test server. @Brutus5000
It has to be considered, that these kind of public markdown files are used in other places (e.g. a forum post or github which both also accept markdown). In these cases we need an absolute url. 

@FredrikBergelin Can you add a screenshot of what even has changed?

FredrikBergelin:
Capture

src/backend/routes/views/staticMarkdownRouter.js
FredrikBergelin, line 1: Even though this became unused after reverting, it might prove useful so I left it in. It allows us to inject a class to the container so each page can be styled individually.

FredrikBergelin:
@Brutus5000 @Sheikah45 @Garanas Merry Christmas! The final texts have just been added including the new moderation page. I also made sure the markdown pages are shown over snowflakes for readability. 

If any of you have time, would be great to get this up on the test server, then release it possibly before new years. Btw, how to access it?

Sheikah45:
@FredrikBergelin it looks like some linting rules failed if you can take a look at that

FredrikBergelin:
@Sheikah45 That should fix it. It got pushed from the wrong account apparently, not on my own computer.

Issue #559, opened 2024-12-11
High CPU and I/O Rate Usage for Snowflake Animation


magge-faf:
If you browse any part of the website that shows the snowflake animation, the CPU demand and I/O rate is unusual high for that little animation. Happens in Firefox and Edge, tested on PC and Laptop.

e.g.:

https://faforever.com/

In my case: Firefox idles around 0.3-0.8% CPU (9800X3D) even when I browse randomly around. Going to faforever.com, and it instantly goes over 6-8% CPU. This is kinda crazy high for the CPU. On my old Laptop, it is just maxing out one core.

![image](https://github.com/user-attachments/assets/09c7048d-57f9-4922-8369-c5e739e3eb04)
![image](https://github.com/user-attachments/assets/c6fcbc48-8904-4681-9060-6c4c036a116c)

**Edit**: Toke a look with the develop tools. Safe to say it is the snowflakes

![image](https://github.com/user-attachments/assets/abc65d59-7eca-4de8-ace7-5f2e4d524a65)

PR #563, opened 2025-01-11
Fixes for the Rules page update

FredrikBergelin:
- Removed snowflakes, see https://github.com/FAForever/website/issues/559 and also they were distracting. We could add a new package that is more stable, but should reasonably be between 24-31 dec, not dec-feb as is now.
- Added in menu items for mobile, was missing (Rules, Moderation)
- Updates to the moderation page texts and added collapsed answers to FAQ.


FredrikBergelin:
@Sheikah45 🙏

Sheikah45 (approved):

PR #564, opened 2025-01-16
Add analytics via Plausible

Garanas:
Adds a reference to [Plausible](https://plausible.io) to provide privacy-friendly analytics. See also the corresponding dashboard that is public:

- https://plausible.io/faforever.com

Note: the dashboard won't work until this pull request is merged 😄 !

The script tag has a [Subresource integrity hash (SRI)](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity). This is generated via:

- https://www.srihash.org/

This opens up various interesting capabilities. As an example, in the future we can introduce [funnels](https://plausible.io/docs/funnel-analysis) to see how people experience the onboarding process.

### DIFF
--- a/src/backend/templates/layouts/default.pug
+++ b/src/backend/templates/layouts/default.pug
@@ -16,7 +16,7 @@ html(lang='en')
         )
 
         //- Analytics via Plausible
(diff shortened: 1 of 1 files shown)


Garanas:
Panic, ha. I put in the wrong domain and tried to hide it. But the force pushes show up here, of course 😄 .

Brutus5000 (approved):

PR #574, opened 2025-07-11
Update privacy policy

Brutus5000:
Drop foreign policies, as they aren't official and translators are everywhere now.


src/backend/templates/views/markdown/privacy.md
```diff
@@ -37,16 +27,51 @@ When you create a new account, you agree to our collection of this User Personal
  - We will use your email address to communicate with you for the purposes of marketing, if you’ve said that’s okay.
  - We will use your email address to communicate with you about important updates to our service.
  - We use your User Personal Information for internal purposes, such as moderation, maintaining logs for security purposes and for legal documentation.
+   - We retain certain technical identifiers (such as IP addresses and device/browser fingerprints) and 3rd party identifiers (such as SteamID or gog.com ID) where necessary to detect repeat abuse, prevent ban evasion, or investigate serious violations of our Terms of Service. This is based on our legitimate interest in maintaining the integrity and security of our community. These identifiers are not used for profiling, advertising, or user behavior analysis unrelated to security or moderation.
```
Garanas, line 30: How do deleted users get processed in analysis such as:

- https://forum.faforever.com/topic/9118/faf-statistics-megathread-2-statistics-boogaloo/2

They end up being anonymized, but still recognizable between data samples?

src/backend/templates/views/markdown/privacy.md
```diff
@@ -37,16 +27,51 @@ When you create a new account, you agree to our collection of this User Personal
...
+```
+
+## Data Retention
+
+We retain personal data only as long as necessary to fulfill the purposes described or to comply with legal obligations. Specifically:
+- Account and profile data are retained while your account is active or until you request deletion.
+- Technical identifiers are stored up to 5 years after account termination
+- Game license ownership related data are stored up to 5 years after account termination
+- Log files and connection data (including IP addresses) are stored for up to 2 years.
+- Backups of data are kept for up to 2 years before being deleted or anonymized.
```
Garanas, line 60: Is the deletion of this data after the time expires automated?

src/backend/templates/views/markdown/privacy.md
```diff
@@ -37,16 +27,51 @@ When you create a new account, you agree to our collection of this User Personal
...
+
+## Your Rights
+
+Under the GDPR, you have the following rights concerning your personal data:
+- **Right of access:** You may request confirmation of whether we process your data and obtain a copy of your data (Art. 15 GDPR).
+- **Right to rectification:** You may request correction of inaccurate or incomplete data (Art. 16 GDPR).
```
Garanas, line 66: This is quite broad, what does this mean in practice? What would be an example of such a request?

Brutus5000: Usually this refers to things like name, address, phone number or things like that. Nothing we actually collect, so this is more of a "we need to inform the user even if nothing of relevance can be changed"

src/backend/templates/views/markdown/privacy.md
Brutus5000: Partially. Some we do manually.

src/backend/templates/views/markdown/privacy.md
Brutus5000: We cannot delete them from the database technically. We could hide them from API querying though, but that needs some extra work.

Brutus5000: No, correction we can't even do that. It would break queries with other players involved. So this data will stay and we are allowed to keep it if other users are involved.

PR #580, opened 2025-08-17
Remove remains of registration and password reset

### LINKED ISSUE #579: [Bug] Register button on /play links to wrong URL
**Bug**:

The "Register" button on https://faforever.com/play currently points to:

`https://faforever.com/account/register` which results in 404

It should instead link to:

`https://user.faforever.com/register`

**Impact / Urgency:**

This prevents new users from registering when they click the "Play" button from the main page. While registration via the top-right user menu still works, the average user will naturally follow the landing-page "Play" button and encounter a broken path.

**Reproduce:**

1. Hit https://faforever.com/play
2. Click on "Register" Button
3. 404 Error

Brutus5000:
Fixes #579


magge-faf:
Awesome! Thank you very much.

PR #595, opened 2026-05-26
Replace ad-hoc cron Scheduler with chained-timeout helper

Brutus5000:
## Summary
- New tiny helpers in `src/backend/cron-jobs/`: `scheduleRepeating` (chained `setTimeout`, can't overlap, returns a `.stop()` handle) and `warmAll` (collects per-action failures via `Promise.allSettled`, logs each, then throws so the scheduler-level log records the failure).
- Single `CACHE_WARM_INTERVAL_MS` constant in `constants.js` replaces the magic `60 * 59 * 1000` that was duplicated in each crawler. Comment explains the relationship to the 60-minute service cache TTL.
- The three crawlers (`leaderboard`, `wordpress`, `clan`) shrink to ~15 lines each — just a list of named warmup actions.
- `AppKernel.startCronJobs` now registers `SIGTERM`/`SIGINT` handlers that call `.stop()` on every scheduler, so the pending timer is cleared on shutdown (no leaked timers under nodemon or Docker stop).
- Deletes the EventEmitter-based `services/Scheduler.js` (22 lines wrapping `setInterval` for no real benefit).

Net diff: **+116 / -196** across 8 files; `cron-jobs/` directory goes from ~190 LOC to ~110.

## Why
The previous `Scheduler` class added indirection without value, and each crawler reimplemented the same boilerplate (per-call success/error handlers, magic interval constant, top-level catch). Three concrete bugs/risks the cleanup also fixes:
1. `setInterval` could schedule a second warmup before the previous one finished — now impossible because the next timer is only set in `finally` after the run completes.
2. No clean shutdown — timer handles were never cleared on signal, so nodemon restarts and Docker `stop` would leak callbacks until the process exited.
3. Per-action failure isolation was copy-pasted into every crawler; now lives in one helper and behaves consistently.

This deliberately does **not** introduce `node-cron` or a job queue (BullMQ etc.) — neither would solve a real problem here, and both would add deps.

## Test plan
- [x] `yarn lint` clean
- [x] `yarn test` — 71/71 (tests don't exercise `startCronJobs`, but they do boot AppKernel and would catch import-level breakage)
- [ ] Manual: deploy to test env, confirm `[scheduler] leaderboard-cache ok in Xms` (or `failed`) lines appear at startup and roughly every 59 minutes
- [ ] Manual: `docker stop` the running container, confirm the process exits cleanly (no lingering timer warnings)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Improved shutdown behavior: background tasks now properly stop when the application receives shutdown signals.

* **Refactor**
  * Optimized cache warming operations to run concurrently instead of sequentially, improving overall performance.
  * Introduced configurable cache refresh intervals with intelligent timing designed to refresh cached data before expiry.

[Review Change Stack](https://app.coderabbit.ai/change-stack/FAForever/website/pull/595?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Brutus5000:
@coderabbitai review

